PR #6, opened 2021-01-30
Add LICENSE.TXT

kilograham:


liamfraser (approved):

LICENSE.TXT
```diff
@@ -0,0 +1,21 @@
+Copyright 2020 (c) 2020 Raspberry Pi (Trading) Ltd.
```
lurch, line 1: 2021 ?

kilograham: seems to be what all the others have; go figure

lurch: Oh, so they do! I never noticed that, oops.

PR #16, opened 2021-03-08
Add boot2 to build information

kilograham:


main.cpp
```diff
@@ -1434,6 +1435,7 @@ void info_guts(memory_access &raw_access) {
                 select_group(build_info);
                 info_pair("sdk version", sdk_version);
                 info_pair("pico_board", pico_board);
+                info_pair("boot2_name", boot2_name);
```
lurch, line 1438: Any particular reason for there being an underscore in `"pico_board"` and `"boot2_name"`, but not in `"sdk version"`, `"build date"`or `"build attributes"` ?

Issue #14, opened 2021-03-06
Support commands other than "reset" for non BOOTSEL device with STDIO USB reset interface


kilograham:
If a device is not in BOOTSEL mode but is exposing a reset interface along with STDIO USB, then we can potentially run any command by first resetting the device to BOOTSEL mode, and then performing the original intended command.

- We should use the same `-f, --force` or whatever from #13 
- We can reset the device into PICOBOOT only mode when doing this to avoid the device ever showing up as a drive

Open question; if we are performing an operation other than "load -x", what should we do after - the device will be left not running. We can either reset it or expose the USB interface at that point (in which case we may as well expose it as soon as reset if we won't be resetting again after the command)

lurch:
> what should we do after

Perhaps another flag to specify the post-action?
Or maybe trying to collapse a `reset; do_command; reset;` sequence into a single `do_command -f;` sequence would be simply better left as separate steps? :shrug:  Perhaps the user would be confused by `do_command -f;` "suddenly" failing if the binary running on the Pico doesn't happen to have the reset vendor interface exposed, whereas just the `reset` command failing _might_ be more obvious?

What might be *even* nicer would be a wrapper-script that can do the same actions either via picotool or SWD, but that's probably a discussion for a later day... (and might be something that the community cooks up anyway?)

Issue #13, opened 2021-03-06
Support "reset" command for non BOOTSEL device with STDIO USB reset interface


kilograham:
SDK 1.1.0 adds (optional) support for code exposing STDIO USB to be reset via another USB vendor interface. This obviously makes sense for the the "reset" command, though we should probably add a `-f, --force` or whatever option to make this happen.

lurch:
IIRC that new USB vendor interface allows resetting to BOOTSEL or resetting to flash boot, so perhaps picotool should expose both those options?

kilograham:
picotool already does

lurch:
I guess that must be only in the Graham-version? https://github.com/raspberrypi/picotool/commits/master

kilograham:
the `picotool reboot` already supports options to specify which type of boot you want (flash boot or usb boot), has since the get go; it is in master

lurch:
Ahh, I was getting confused by the already-existing-reboot (i.e. over PICOBOOT?) and the new-USB-vendor-reboot.
Also, I'd never thought to run `picotool help reboot` which is the **only** place where this is ever mentioned :stuck_out_tongue_winking_eye:

lurch:
And based on our internal discussions, I wonder if `picotool reboot` also ought to have an option for "reboot into BOOTSEL mode but only enable PICOBOOT, don't enable the MSD interface" ? :shrug:

PR #23, opened 2021-04-20
Add ability to access device using stdio_usb

kilograham:


main.cpp
```diff
@@ -262,7 +267,9 @@ auto device_selection =
         (option("--bus") & integer("bus").min_value(1).max_value(255).set(settings.bus)
             .if_missing([] { return "missing bus number"; })) % "Filter devices by USB bus number" +
         (option("--address") & integer("addr").min_value(1).max_value(127).set(settings.address)
-            .if_missing([] { return "missing address"; })) % "Filter devices by USB device address"
+            .if_missing([] { return "missing address"; })) % "Filter devices by USB device address" +
+        option('f', "--force").set(settings.force) % "Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. Unless the command itself causes a reboot, the device will be rebooted back to application mode" +
+                option('F', "--force-no-reboot").set(settings.no_reboot_if_forced) % "Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. Unless the command itslef causes a reboot, the device will be left connected and accessible to picotool, but without the RPI-RP2 drive mounted"
```
lurch, line 272: typo: itslef -> itself

main.cpp
```diff
@@ -444,6 +457,16 @@ std::basic_string<T> uppercase(const std::basic_string<T>& s)
 
 clipp::formatting_ostream<std::ostream> fos(std::cout);
 
+static void sleep_ms(int ms) {
+#ifdef __unix__
```
lurch, line 461: This define doesn't work on MacOSX.
https://stackoverflow.com/a/16107549 suggests that it ought to be:
```
#if defined(__unix__) || defined(__unix) || \
        (defined(__APPLE__) && defined(__MACH__))
```
but I guess it probably makes more sense to stay consistent with lines 2019 and 2020 and use:
```
#if defined(__linux__) || defined(__APPLE__)
```
?

P.S. Just out of curiosity I checked the `pico-sdk` repo, and that uses an inconsistent mixture of methods:
```
$ git grep "linux\|unix\|APPLE"
src/host/hardware_timer/timer.c:#if defined(__unix__) || defined(__APPLE__)
src/host/hardware_timer/timer.c:#if defined(__unix__) || defined(__APPLE__)
src/host/hardware_timer/timer.c:#if defined(__unix__) || defined(__APPLE__)
src/host/hardware_timer/timer.c:#if defined(__unix__)
src/host/hardware_uart/uart.c:#if defined(__unix) || defined(__APPLE__)
src/host/pico_platform/include/pico/platform.h:#ifdef __unix__
```

kilograham: the `timer.c` ones are correct; the `platform.h` one seems fine, because it seems to be include anyway on apple (or maybe by Clang)

kilograham: i fixed the macos build

main.cpp
```diff
@@ -1213,6 +1236,7 @@ void build_rmap_uf2(FILE *file, range_map<size_t>& rmap) {
             block.magic_end == UF2_MAGIC_END) {
             if (block.flags & UF2_FLAG_FAMILY_ID_PRESENT && block.file_size == RP2040_FAMILY_ID &&
                 !(block.flags & UF2_FLAG_NOT_MAIN_FLASH) && block.payload_size == PAGE_SIZE) {
+                printf("%08x\n", block.target_addr);
```
lurch, line 1239: Is this a debugging-print which you forgot to remove?

kilograham: yes

lurch:
If I use the `-f` or `-F` options on an RP2040 which is _already_ in BOOTSEL mode, I get an error saying "ERROR: Forced command requires a single rebootable RP2040 device to be targeted." which could be slightly confusing. Maybe if the RP2040 is already in BOOTSEL mode it should ignore any 'force' flags?

kilograham:
> If I use the `-f` or `-F` options on an RP2040 which is _already_ in BOOTSEL mode, I get an error saying "ERROR: Forced command requires a single rebootable RP2040 device to be targeted." which could be slightly confusing. Maybe if the RP2040 is already in BOOTSEL mode it should ignore any 'force' flags?

Good point; `-f` should be "reboot it if you have to"

lurch:
When you've got multiple RP2040 devices plugged in at once (and so you need to use the `--address` option to specify which one you're targeting), the 'force' option becomes kinda confusing. This is because when the device reboots, it enumerates with a different address, and so the command that you were trying to run then reports "ERROR: Communication with RP2040 device failed".
It pains me to suggest removing functionality, but maybe the easiest way to solve this is to drop the `-F` flag (which IMHO is confusing anyway (but maybe that's just the help-string)), and have `reboot` be the _only_ command that accepts a `-f` flag? And then once the device has been force-rebooted the user can work out what the device's new address is, and run `picotool` a second time to run the command they actually want?
A more complex solution might be to disallow the `-f` or `-F` flags being used with any command that _isn't_ `reboot` when you're also specifying the `--address` option? :shrug: 

---

Also, I wonder if a `pictool list` command might be useful? Consider the following scenario:
```
% lsusb | grep 2e8a                                       
Bus 020 Device 012: ID 2e8a:000a 2e8a Pico  Serial: 000000000000
Bus 020 Device 009: ID 2e8a:0005 2e8a Board in FS mode  Serial: 000000000000
% ./picotool info

but:

Device at bus 20, address 9 appears to be a RP2040 MicroPython device not in BOOTSEL mode.
Device at bus 20, address 12 appears to be a RP2040 device with a USB serial connection, so consider -f or -F.
```
if I then plug in a Pico with a blank Flash chip (wiped using `flash_nuke.uf2`) I get:
```
% lsusb | grep 2e8a
Bus 020 Device 012: ID 2e8a:000a 2e8a Pico  Serial: 000000000000
Bus 020 Device 009: ID 2e8a:0005 2e8a Board in FS mode  Serial: 000000000000
Bus 020 Device 013: ID 2e8a:0003 2e8a RP2 Boot  Serial: E0C912D24340
% ./picotool info
Program Information
 none
```
In this latter case it would be nice if e.g. a `./picotool list` could tell me:
```
Device at bus 20, address 9 appears to be a RP2040 MicroPython device not in BOOTSEL mode.
Device at bus 20, address 12 appears to be a RP2040 device with a USB serial connection, so consider -f or -F.
Device at bus 20, address 13 appears to be a RP2040 device in BOOTSEL mode.
```
For bonus points, it could even include the other VID/PID combos from https://github.com/raspberrypi/usb-pid ? :wink: 

---

And finally (for now) it would be nice if picotool was able to tell the difference between a "RP2040 device with a USB serial connection" where the reset vendor interface is active (so the `-f` flag will (probably?) work), and a "RP2040 device with a USB serial connection" where the reset vendor interface **isn't** active (so the `-f` flag (definitely?) won't work). I _think_ that information might be available from the USB descriptor, but I've obviously got no idea how tricky it'd be to add that support to picotool.

main.cpp
```diff
@@ -2027,15 +2104,73 @@ int main(int argc, char **argv) {
...
+                        settings.force = false;
+                        settings.address = -1;
+                        settings.bus = -1;
+                        continue;
+                    }
+                }
+                selected_cmd->execute(devices);
+                if (tries) {
+                    if (settings.no_reboot_if_forced) {
+                        printf("The device has been left accessible, but without the drive mounted; use 'picotool reboot' to reboot into regular BOOTSEL mode or application mode.");
```
lurch, line 2162: Why does this use `printf` when `cout` is used elsewhere? Also, is the trailing `\n` missing?

kilograham:
> When you've got multiple RP2040 devices plugged in at once (and so you need to use the `--address` option to specify which one you're targeting), the 'force' option becomes kinda confusing. This is because when the device reboots, it enumerates with a different address, and so the command that you were trying to run then reports "ERROR: Communication with RP2040 device failed".

Yes, this is a (known) problem... sadly libusb on Windows does not support the _good_ way to handle this (i.e. the ability to watch for devices appearing/disappearing)

> It pains me to suggest removing functionality, but maybe the easiest way to solve this is to drop the `-F` flag (which IMHO is confusing anyway (but maybe that's just the help-string)), and have `reboot` be the _only_ command that accepts a `-f` flag? And then once the device has been force-rebooted the user can work out what the device's new address is, and run `picotool` a second time to run the command they actually want?

We do not want to throw the baby out with the bath water. The single Pico case is likely to be common and an integrated`picotool -f` with all sensible commands means the ability to use in scripts etc where user intervention is not desriable.

We should try to improve the multiple Pico case in the future (we can probably do a _diff_ of the set of USB descriptors to figure out which device was rebooted in the majority of cases)

> A more complex solution might be to disallow the `-f` or `-F` flags being used with any command that _isn't_ `reboot` when you're also specifying the `--address` option? 🤷

That seems reasonable

> 
> Also, I wonder if a `pictool list` command might be useful? Consider the following scenario:
> 
> ```
> % lsusb | grep 2e8a                                       
> Bus 020 Device 012: ID 2e8a:000a 2e8a Pico  Serial: 000000000000
> Bus 020 Device 009: ID 2e8a:0005 2e8a Board in FS mode  Serial: 000000000000
> % ./picotool info
> No accessible RP2040 devices in BOOTSEL mode were found.
> 
> but:
> 
> Device at bus 20, address 9 appears to be a RP2040 MicroPython device not in BOOTSEL mode.
> Device at bus 20, address 12 appears to be a RP2040 device with a USB serial connection, so consider -f or -F.
> ```
> 
> if I then plug in a Pico with a blank Flash chip (wiped using `flash_nuke.uf2`) I get:
> 
> ```
> % lsusb | grep 2e8a
> Bus 020 Device 012: ID 2e8a:000a 2e8a Pico  Serial: 000000000000
> Bus 020 Device 009: ID 2e8a:0005 2e8a Board in FS mode  Serial: 000000000000
> Bus 020 Device 013: ID 2e8a:0003 2e8a RP2 Boot  Serial: E0C912D24340
> % ./picotool info
> Program Information
>  none
> ```
> 
> In this latter case it would be nice if e.g. a `./picotool list` could tell me:
> 
> ```
> Device at bus 20, address 9 appears to be a RP2040 MicroPython device not in BOOTSEL mode.
> Device at bus 20, address 12 appears to be a RP2040 device with a USB serial connection, so consider -f or -F.
> Device at bus 20, address 13 appears to be a RP2040 device in BOOTSEL mode.
> ```
> 
> For bonus points, it could even include the other VID/PID combos from https://github.com/raspberrypi/usb-pid ? 😉
> 
> And finally (for now) it would be nice if picotool was able to tell the difference between a "RP2040 device with a USB serial connection" where the reset vendor interface is active (so the `-f` flag will (probably?) work), and a "RP2040 device with a USB serial connection" where the reset vendor interface **isn't** active (so the `-f` flag (definitely?) won't work). I _think_ that information might be available from the USB descriptor, but I've obviously got no idea how tricky it'd be to add that support to picotool.

All of this should be discussed in a separate issue

main.cpp
```diff
@@ -1908,33 +1930,84 @@ void verify_command::execute(device_map &devices) {
     }
 }
 
+static int reboot_device(libusb_device *device, bool bootsel, uint disable_mask=0) {
+    // ok, the device isn't in USB boot mode, let's try to reboot via baud rate hack
```
lurch, line 1934: Is this comment correct? Should it say "reboot via reset vendor interface" instead?
Should this PR _also_ include the baud-rate-hack, or will that be added in a later PR?

kilograham: it should say that; resetting by baud is not supported, as it can't be done via libusb when the device is claimed by a comms driver, and finding the comms device is complex & platform dependent

liamfraser (dismissed):

lurch:
For cross-linking purposes: I believe this fixes both #13 and #14 and links to https://github.com/raspberrypi/pico-sdk/pull/197

kilograham:
yes, i bet that is it

main.cpp
lurch: Yep, this is definitely missing a `\n`...
image

lurch:
> If I use the -f or -F options on an RP2040 which is already in BOOTSEL mode, I get an error saying "ERROR: Forced command requires a single rebootable RP2040 device to be targeted." which could be slightly confusing. Maybe if the RP2040 is already in BOOTSEL mode it should ignore any 'force' flags?

I'm still seeing this behaviour on Linux - your previous reply implied this might be changing?

main.cpp
```diff
@@ -263,6 +269,10 @@ auto device_selection =
             .if_missing([] { return "missing bus number"; })) % "Filter devices by USB bus number" +
         (option("--address") & integer("addr").min_value(1).max_value(127).set(settings.address)
             .if_missing([] { return "missing address"; })) % "Filter devices by USB device address"
+#if !defined(_WIN32)
+        + option('f', "--force").set(settings.force) % "Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. Unless the command itself causes a reboot, the device will be rebooted back to application mode" +
+        option('F', "--force-no-reboot").set(settings.no_reboot_if_forced) % "Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. Unless the command itself causes a reboot, the device will be left connected and accessible to picotool, but without the RPI-RP2 drive mounted"
```
lurch, line 274: FWIW I still find this "Unless the command itself causes a reboot" text quite confusing - so other users might too? :man_shrugging:

lurch:
If a Pico is in BOOTSEL mode, and I run `sudo ./picotool reboot -u` it successfully reboots back into BOOTSEL mode, but it also prints `ERROR: Communication with RP2040 device failed` which seems a bit odd?

kilograham:
can you include platform you are running on with the issues

lurch:
Dunno if it's deliberate behaviour or not, but on Linux when I run `picotool help reboot` the `-f` option is listed under the `Selecting the device to reboot` heading, whereas on Windows when I run the same command the `-f` option appears under the `Reboot type` heading. (The `-F` option is also missing on Windows, but I realise that's intentional)

main.cpp
kilograham: > If a Pico is in BOOTSEL mode, and I run `sudo ./picotool reboot -u` it successfully reboots back into BOOTSEL mode, but it also prints `ERROR: Communication with RP2040 device failed` which seems a bit odd?

a yes, i forgot - lost in the mists of time; will fix

> > If I use the -f or -F options on an RP2040 which is already in BOOTSEL mode, I get an error saying "ERROR: Forced command requires a single rebootable RP2040 device to be targeted." which could be slightly confusing. Maybe if the RP2040 is already in BOOTSEL mode it should ignore any 'force' flags?
> 
> I'm still seeing this behaviour on Linux - your previous reply implied this might be changing?

ah yes, i forgot

lurch:
On Windows, if I run `picotool reboot -f -a -u` it reboots into BOOTSEL mode, but if I run `picotool reboot -f -u -a` it reboots to application mode!
It should probably just error if you specify both the `-a` and the `-u` flags?

Ahh, same behaviour on Linux too.

main.cpp
kilograham: > FWIW I still find this "Unless the command itself causes a reboot" text quite confusing - so other users might too? man_shrugging

I have changed to "After executing the command (unless the command itself is a 'reboot')"

kilograham:
> On Windows, if I run `picotool reboot -f -a -u` it reboots into BOOTSEL mode, but if I run `picotool reboot -f -u -a` it reboots to application mode!
> It should probably just error if you specify both the `-a` and the `-u` flags?
> 
> Ahh, same behaviour on Linux too.

will fix this

lurch:
The Mac build seems to be broken again? :confused: 
```
[ 25%] Building CXX object CMakeFiles/picotool.dir/main.cpp.o
/Users/andrew/pico/picotool/main.cpp:472:5: error: use of undeclared identifier 'Sleep'
    Sleep(ms);
    ^
1 error generated.
make[2]: *** [CMakeFiles/picotool.dir/main.cpp.o] Error 1
```

main.cpp
```diff
@@ -1908,33 +1935,84 @@ void verify_command::execute(device_map &devices) {
     }
 }
 
+static int reboot_device(libusb_device *device, bool bootsel, uint disable_mask=0) {
+    // ok, the device isn't in USB boot mode, let's try to reboot via vender interface
```
lurch, line 1939: vender should be vendor :wink:

lurch:
On Linux x64, the "info <filename>" seems to be broken?
If I do
```
./picotool info ../../pico-examples/build/blink/blink.uf2
```
it just hangs (doesn't do anything), and when I Ctrl-C I get `Segmentation fault (core dumped)` :slightly_frowning_face:

lurch:
On Linux x64, if I do
```
sudo ./picotool reboot -a -u
```
it now reports:
```
ERROR: Cannot specify both -u and -a reboot options
Segmentation fault
```
The first line is good, the second line is bad :wink: 

But I can confirm the "ignore -f / -F flag if  specified but not needed" issue seems to have been fixed :tada: 

Hmmm, if I do multiple
```
sudo ./picotool reboot -u
```
in a row, it sometimes works with no error message, but sometimes it prints `ERROR: Communication with RP2040 device failed` and sometimes prints `ERROR: The RP2040 device returned an error: <unknown>`.

Doing multiple
```
sudo ./picotool reboot -a -f
```
or
```
sudo ./picotool reboot -a -F
```
in a row always seems to work with no errors reported. :+1: 

I dunno if it's worth "fixing", but these commands work:
```
sudo ./picotool reboot -a -f -F
sudo ./picotool reboot -a -F -f
sudo ./picotool reboot -f -F -a
sudo ./picotool reboot -F -f -a
```
but
```
sudo ./picotool reboot -f -a -F
```
reports:
```
ERROR: unexpected option: -F

SYNOPSYS:
    picotool reboot [-a] [-u] [--bus <bus>] [--address <addr>] [-f] [-F]

Use "picotool help reboot" for more info
```
and
```
sudo ./picotool reboot -F -a -f
```
reports:
```
ERROR: unexpected option: -f

SYNOPSYS:
    picotool reboot [-a] [-u] [--bus <bus>] [--address <addr>] [-f] [-F]

Use "picotool help reboot" for more info
```

lurch:
On Windows 10, if I do
```
picotool -a -u
```
I get:
image

I noticed that on Linux I get a nice "The device was rebooted into _[application/BOOTSEL]_ mode" if I run "sudo picotool reboot _[-a/-u]_ -f", but there's no message printed when I run `picotool reboot -a -f` or `picotool reboot -u -f` on Windows?

Windows sometimes also prints `ERROR: Communication with RP2040 device failed` after running `picotool reboot -u -f`.

I dunno if it's worth special-casing the error-message, but if the Pico is in application mode (on Windows) and I run
```
picotool reboot -a
```
it tells me:
```

but:

Device at bus 2, address 23 appears to be a RP2040 device with a USB serial connection, not in BOOTSEL mode. You can force reset it into BOOTSEL mode via 'picotool reboot -f -u' first.
```
however it seems that `picotool reboot -a -f` would work in this scenario?

main.cpp
kilograham: fixing

kilograham:
> it

hmm i do not see this will re-push shortly to make sure everything is in sync

kilograham:
> On Linux x64, the "info <filename>" seems to be broken?
> If I do
> 
> ```
> ./picotool info ../../pico-examples/build/blink/blink.uf2
> ```
> 
> it just hangs (doesn't do anything), and when I Ctrl-C I get `Segmentation fault (core dumped)` slightly_frowning_face

uninitialized variable bug was introduced, will push shortly

kilograham:
as for the -F -a -f stuff, i wrote a whole new command line parser for this tool, and there are a few edge cases... trying to fix them without breaking other stuff is not worth it. (it does work if you put the options in the right place after all)

kilograham:
@lurch can you give this a final once over, i want to merge

CMakeLists.txt
```diff
@@ -30,6 +30,16 @@ else()
     add_subdirectory(${PICO_SDK_PATH}/src/host/pico_platform pico_platform)
 
     add_executable(picotool main.cpp)
+    set(PICOTOOL_VERSION 1.1.0-develop)
+    set(SYSTEM_VERSION "${CMAKE_SYSTEM_NAME} ${CMAKE_SYSTEM_VERSION}")
+    set(COMPILER_INFO "${CMAKE_C_COMPILER_ID}-${CMAKE_C_COMPILER_VERSION} ${CMAKE_BUILD_TYPE}")
```
lurch, line 35: When I run `./picotool version` (on Linux) I get `picotool v1.1.0-develop (Linux 5.4.0-77-generic, GNU-7.5.0 )` - does this mean that the `${CMAKE_BUILD_TYPE}` variable needs to have a default value given to it, if not explicitly set?

kilograham: ah, i guess i always build from my idea which sets it explicitly

lurch:
(Testing on a Pi400, but I wouldn't expect other platforms to behave differently...)

There seems to be an odd interaction between `picotool load -x` and the 'force' flags?
```
$ sudo ./picotool load -x p4.uf2 -f
The device was rebooted as requested into BOOTSEL mode so the command can be executed.

Loading into Flash: [==============================]  100%
The device has been left accessible, but without the drive mounted; use 'picotool reboot' to reboot into regular BOOTSEL mode or application mode.
$ sudo ./picotool load -x p4.uf2 -F
The device was rebooted as requested into BOOTSEL mode so the command can be executed.

Loading into Flash: [==============================]  100%
The device has been left accessible, but without the drive mounted; use 'picotool reboot' to reboot into regular BOOTSEL mode or application mode.
```

Note how the `-f` and `-F` flags are behaving identically, which I don't think is right? :confused:

kilograham:
> (Testing on a Pi400, but I wouldn't expect other platforms to behave differently...)
> 
> There seems to be an odd interaction between `picotool load -x` and the 'force' flags?
> 
> ```
> $ sudo ./picotool load -x p4.uf2 -f
> The device was rebooted as requested into BOOTSEL mode so the command can be executed.
> 
> Loading into Flash: [==============================]  100%
> The device has been left accessible, but without the drive mounted; use 'picotool reboot' to reboot into regular BOOTSEL mode or application mode.
> $ sudo ./picotool load -x p4.uf2 -F
> The device was rebooted as requested into BOOTSEL mode so the command can be executed.
> 
> Loading into Flash: [==============================]  100%
> The device has been left accessible, but without the drive mounted; use 'picotool reboot' to reboot into regular BOOTSEL mode or application mode.
> ```
> 
> Note how the `-f` and `-F` flags are behaving identically, which I don't think is right? 😕

yeah the main issue is that `-x` causes a reboot so should be considered similarly to a reboot command

lurch:
I built and tested the latest version this afternoon on Linux x64, Raspberry Pi OS (on a Pi400), Windows 10 x64 and MacOS Catalina x64. Apart from my "little catches" mentioned earlier, it all seems to work pretty well. (with the obvious difference that it's _only_ the 'reboot' command on Windows which accepts a 'force' flag)

However I do wonder if it wouldn't be simpler / more consistent to simply make:
```
        -F, --force-no-reboot
            Force a device not in BOOTSEL mode but running compatible code to reset
            so the command can be executed. After executing the command (unless the
            command itself is a 'reboot') the device will be left connected and
            accessible to picotool, but without the RPI-RP2 drive mounted
```
behave the same as `picotool reboot -u` i.e. have the `-F` option leave the RP2040 in "full BOOTSEL mode", rather than this unusual "the device will be left connected and accessible to picotool, but without the RPI-RP2 drive mounted" mode? Unless there's a good reason for introducing this latter mode? :shrug:

lurch:
Oh, and I always need to apply the change from #27 before I'm able to build on Windows.

CMakeLists.txt
kilograham: fixed

kilograham:
> I built and tested the latest version this afternoon on Linux x64, Raspberry Pi OS (on a Pi400), Windows 10 x64 and MacOS Catalina x64. Apart from my "little catches" mentioned earlier, it all seems to work pretty well +1 (with the obvious difference that it's _only_ the 'reboot' command on Windows which accepts a 'force' flag)
> 
> However I do wonder if it wouldn't be simpler / more consistent to simply make:
> 
> ```
>         -F, --force-no-reboot
>             Force a device not in BOOTSEL mode but running compatible code to reset
>             so the command can be executed. After executing the command (unless the
>             command itself is a 'reboot') the device will be left connected and
>             accessible to picotool, but without the RPI-RP2 drive mounted
> ```
> 
> behave the same (on Linux and Mac) as `picotool reboot -u` i.e. have the `-F` option leave the RP2040 in "full BOOTSEL mode", rather than this unusual "the device will be left connected and accessible to picotool, but without the RPI-RP2 drive mounted" mode? Unless there's a good reason for introducing this latter mode? shrug

yeah this is my preferred behavior... `-f` is preferable because you don't end up with the drive appearing as USB for no reason, so I'd like to keep it, confusing as it may be

kilograham:
> Oh, and I always need to apply the change from #27 before I'm able to build on Windows.

i'm not 100% convinced that it is a good general fix, but i have included it (here now), because it is probably not worse than what was there.

kilograham:
I made a bunch of subtle changes to the wording of messages, so hopefully things make more sense now.

If does seem possible to get the Ubuntu USB stack confused (apparently) in so far as the device can be back in USBBOOT mode, but the OS still thinks it is connected with CDC available. (This seems to be the cause of -f saying it is rebooting, then the device not being found anyway... i have made the error messages better in this eventuality)

main.cpp
```diff
@@ -399,19 +417,53 @@ struct help_command : public cmd {
     }
 } help_cmd;
 
+struct version_command : public cmd {
+    version_command() : cmd("version") {}
+    bool execute(device_map &devices) override {
+        if (settings.version.semantic)
+            std::cout << PICOTOOL_VERSION << "\n";
+        else
+            std::cout << "picotool v" << PICOTOOL_VERSION << " (" << SYSTEM_VERSION << ", " << COMPILER_INFO << ")\n";
```
lurch, line 426: Only a small nit-pick, but this displays e.g.
```
picotool v1.1.0-develop (Linux 5.4.0-81-generic, GNU-7.5.0 Release)
```
_Maybe_ there should be an additional comma between `GNU-7.5.0` and `Release` ?

lurch:
Running the latest version on x64 Linux, I get this sequence of commands:
```bash
$ ./picotool info

but:

Device at bus 3, address 46 appears to be a RP2040 device with a USB serial
    connection, so consider -f (or -F) to force reboot in order to run the
    command.
$ ./picotool info -F
ERROR: Unable to access device to reboot it; Use sudo or setup a udev rule

$ sudo ./picotool info -F
The device was asked to reboot into BOOTSEL mode so the command can be executed.

Program Information
 name:      hello_usb
 web site:  https://github.com/raspberrypi/pico-examples/tree/HEAD/hello_world/usb
 features:  USB stdin / stdout

            The device has been left accessible, but without the drive mounted;
            use 'picotool reboot' to reboot into regular BOOTSEL mode or
            application mode.
$ ./picotool info

but:

Device at bus 3, address 49 appears to be a RP2040 device in BOOTSEL mode, but
    picotool was unable to connect
$ sudo ./picotool info
Program Information
 name:      hello_usb
 web site:  https://github.com/raspberrypi/pico-examples/tree/HEAD/hello_world/usb
 features:  USB stdin / stdout
```

Would it perhaps make sense to also display the "Use sudo or setup a udev rule" message as part of the "Device at bus 3, address 49 appears to be a RP2040 device in BOOTSEL mode, but picotool was unable to connect" output in the penultimate command above?
Or does "a RP2040 device in BOOTSEL mode, but picotool was unable to connect" cover more error cases than just "Use sudo or setup a udev rule" ?

lurch:
> the main issue is that -x causes a reboot so should be considered similarly to a reboot command

This appears to be fixed now :+1:  (both `sudo ./picotool load p4.uf2 -x -f` and `sudo ./picotool load p4.uf2 -x -F` display "The device was rebooted to start the application.")
However I wonder if the `(unless the command itself is a 'reboot')` in the `-F, --force-no-reboot` section of the `help` output should now say `(unless the command itself is a 'reboot' or a 'load -x')` ?

lurch:
> yeah this is my preferred behavior... -f is preferable because you don't end up with the drive appearing as USB for no reason, so I'd like to keep it, confusing as it may be

That's fair enough. However there's part of me that wonders if, for the sake of symmetry, there ought to be a dedicated flag to the reboot command (e.g. `-p`) that could also be used to enter this "connected and accessible to picotool, but without the RPI-RP2 drive mounted" state? :shrug:

lurch:
Another minor-nitpick thing; in the output of `picotool help` I see:
```
SYNOPSYS:
    picotool info [-b] [-p] [-d] [-l] [-a] [--bus <bus>] [--address <addr>] [-f]
                [-F]
    picotool info [-b] [-p] [-d] [-l] [-a] <filename> [-t <type>]
    picotool load [-v] [-x] <filename> [-t <type>] [-o <offset>] [--bus <bus>]
                [--address <addr>] [-f] [-F]
    picotool save [-p] [--bus <bus>] [--address <addr>] [-f] [-F] <filename> [-t
                <type>]
    picotool save -a [--bus <bus>] [--address <addr>] [-f] [-F] <filename> [-t
                <type>]
    picotool save -r <from> <to> [--bus <bus>] [--address <addr>] [-f] [-F]
                <filename> [-t <type>]
    picotool verify [--bus <bus>] [--address <addr>] [-f] [-F] <filename> [-t
                <type>] [-r <from> <to>] [-o <offset>]
    picotool reboot [-a] [-u] [--bus <bus>] [--address <addr>] [-f] [-F]
    picotool version [-s]
    picotool help [<cmd>]
```
Looks a bit odd that sometimes the command-options are listed _before_ the `[--bus <bus>] [--address <addr>] [-f] [-F]` part, and at other times they're listed _after_ the `[--bus <bus>] [--address <addr>] [-f] [-F]` part?

lurch:
```
$ ./picotool help save
SAVE:
    Save the program / memory stored in flash on the device to a file.

SYNOPSYS:
    picotool save [-p] [--bus <bus>] [--address <addr>] [-f] [-F] <filename> [-t
                <type>]
    picotool save -a [--bus <bus>] [--address <addr>] [-f] [-F] <filename> [-t
                <type>]
    picotool save -r <from> <to> [--bus <bus>] [--address <addr>] [-f] [-F]
                <filename> [-t <type>]

OPTIONS:
    Selection of data to save
        -p, --program
            Save the installed program only. This is the default
        -a, --all
            Save all of flash memory
        -r, --range
            Save a range of memory. Note that UF2s always store complete 256
            byte-aligned blocks of 256 bytes, and the range is expanded
            accordingly
        <from>
            The lower address bound in hex
        <to>
            The upper address bound in hex
    Source device selection
        --bus <bus>
            Filter devices by USB bus number
        --address <addr>
        -f, --force
            Force a device not in BOOTSEL mode but running compatible code to
            reset so the command can be executed. After executing the command
            (unless the command itself is a 'reboot') the device will be
            rebooted back to application mode
        -F, --force-no-reboot
            Force a device not in BOOTSEL mode but running compatible code to
            reset so the command can be executed. After executing the command
            (unless the command itself is a 'reboot') the device will be left
            connected and accessible to picotool, but without the RPI-RP2 drive
            mounted
    File to save to
        <filename>
            The file name
        -t <type>
            Specify file type (uf2 | elf | bin) explicitly, ignoring file
            extension
$ sudo ./picotool save p3.elf
ERROR: Save to ELF file is not supported
```
Perhaps it shouldn't mention `elf` in the save help-text?

kilograham:
> 
> Perhaps it shouldn't mention `elf` in the save help-text?

this i don't care about - we might support it in the future

lurch:
Not that it matters, but any idea why an out-of-the-box build creates a `Release` build on Linux and macOS, but a `Debug` build on Windows? (according to the output of `picotool version`).

I can confirm that as long as the `PICO_SDK_PATH` and `LIBUSB_ROOT` environment variables are set appropriately, this PR now builds on Windows without any extra tweaks needed. :+1:

main.cpp
kilograham: fixed

PR #27, opened 2021-04-30
Fix FindLIBUSB.cmake for Windows builds

lurch:
Based on a modified version of the suggestions in https://www.raspberrypi.org/forums/viewtopic.php?p=1854531


cmake/FindLIBUSB.cmake
```diff
@@ -19,10 +19,12 @@ else (LIBUSB_INCLUDE_DIR AND LIBUSB_LIBRARIES)
         pkg_check_modules(PC_LIBUSB libusb-1.0)
     ENDIF(NOT WIN32)
     FIND_PATH(LIBUSB_INCLUDE_DIR libusb.h
+            HINTS $ENV{LIBUSB_ROOT}/include/libusb-1.0
             PATHS ${PC_LIBUSB_INCLUDEDIR} ${PC_LIBUSB_INCLUDE_DIRS})
-    FIND_LIBRARY(LIBUSB_LIBRARIES NAMES usb-1.0
+    FIND_LIBRARY(LIBUSB_LIBRARIES NAMES usb usb-1.0 libusb-1.0
+            HINTS $ENV{LIBUSB_ROOT}/VS2019/MS32/static
             PATHS ${PC_LIBUSB_LIBDIR} ${PC_LIBUSB_LIBRARY_DIRS})
```
lurch, line 26: @kilograham Looking at https://cmake.org/cmake/help/latest/command/find_library.html have I actually got `HINTS` and `PATHS` the wrong way around?
i.e. does the usage of `${PC_LIBUSB_LIBDIR}` count as "computed by system introspection" meaning that it ought to use the `HINTS` option; and does the usage of `$ENV{LIBUSB_ROOT}` count as a "hard-coded guess" meaning that it ought to use the `PATHS` option?

lurch:
> fixed via other PR

Fixed via https://github.com/raspberrypi/picotool/commit/757f2aee3c2c988b4d0c9c15907e243350381d0a :wink:

PR #28, opened 2021-05-01
Swap endian-ness in tools on big endian host

zeldin:
This depends on raspberrypi/pico-sdk#329.

### DIFF
--- a/elf.h
+++ b/elf.h
@@ -17,43 +17,44 @@
 
 #define PT_LOAD 0x00000001u
 
+/* Note, only little endian ELFs handled */
 #pragma pack(push, 1)
 struct elf_header {
-    uint32_t    magic;
+    le_uint32_t magic;
     uint8_t     arch_class;
     uint8_t     endianness;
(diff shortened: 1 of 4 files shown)


elf.h
```diff
@@ -17,44 +17,45 @@
 
 #define PT_LOAD 0x00000001u
 
+/* Note, only little endian ELFs handled */
```
lurch, line 20: Would it make sense for `elf2uf2` and `picotool` to both pick up a common `elf.h` header from `pico-sdk`? :shrug: 
(These PRs are a bit over my head, so apologies if that's a silly suggestion.)

josch:
I have https://github.com/raspberrypi/pico-sdk/pull/329 applied to the Debian packaging of pico-sdk and am thus also interested in this patch for picotool.

Can you think of a nice test case for picotool which would *not* require hardware and would fail on big endian machines? Something I can run as part of the Debian autopkgtests to make sure this works?

Thanks!

PR #35, opened 2021-06-20
Add cmake install command

### LINKED ISSUE #4: No rule to install `picotool` binary
All the cmake system is in place except for the installation of `picotool` to `/usr/bin` target.
When `make install` is run, we get:
```
make: *** No rule to make target 'install'.  Stop.
```

markuspi:
Add an install rule to CMakeLists.txt to support `make install`.
Will copy the `picotool` executable to `${CMAKE_INSTALL_PREFIX}/bin/`.
As per [documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html), the install prefix defaults to `/usr/local` on unix and `c:/Program Files/${PROJECT_NAME}` on windows.

A custom prefix can be defined using `cmake -DCMAKE_INSTALL_PREFIX:PATH=/path/to/prefix ..`

Successfully tested on ubuntu.

Resolves #4


kilograham (commented):
Why did you move the code out of the IF?

kilograham:
added via separate PR

PR #41, opened 2021-10-05
add -n option to load to avoid overwriting existing program

kilograham:


lurch:
Just out of curiosity, what's the use-case for this? Won't the RP2040 bootrom always just blindly execute the binary at the default load-address?

lurch:
Is this commit _supposed_ to be in #23 as well? Although 45d182fab96d93e8f48120466acae03ab1140d6b seems to contain more changes than 889469ceea61c29c9632e8654c6e21a12c9ed24f

kilograham:
There are so many overlapping commits that Depend on each other that it is too much work to separate them out; I will continue fixing/rebasing as I merge

PR #46, opened 2022-01-02
add --update option to picotool load to skip unchanged sectors

rlane:
Disabled by default, since reading flash takes about 7% extra time if no sectors can be skipped. For my use case it saves more than 50% time.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -404,7 +404,7 @@ struct load_command : public cmd {
             (
                 option('n', "--no-overwrite").set(settings.load.no_overwrite) % "When writing flash data, do not overwrite an existing program in flash. If picotool cannot determine the size/presence of the program in flash, the command fails" +
                 option('N', "--no-overwrite-unsafe").set(settings.load.no_overwrite_force) % "When writing flash data, do not overwrite an existing program in flash. If picotool cannot determine the size/presence of the program in flash, the load continues anyway" +
-                option('u', "--update").set(settings.load.update) % "Skip flash writes when the data is already present" +
+                option('u', "--update").set(settings.load.update) % "Skip writing flash sectors that already contain identical data" +
                 option('v', "--verify").set(settings.load.verify) % "Verify the data was written correctly" +
                 option('x', "--execute").set(settings.load.execute) % "Attempt to execute the downloaded file as a program after the load"
             ).min(0).doc_non_optional(true) % "Post load actions" +
--- a/main.cpp
+++ b/main.cpp
@@ -278,6 +278,7 @@ struct _settings {
         bool execute = false;
         bool no_overwrite = false;
         bool no_overwrite_force = false;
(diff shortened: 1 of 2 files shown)


main.cpp
```diff
@@ -1859,9 +1861,18 @@ bool load_command::execute(device_map &devices) {
                     file_buf.insert(file_buf.begin(), read_range.from - aligned_range.from, 0);
                     file_buf.insert(file_buf.end(), aligned_range.to - read_range.to, 0);
                     assert(file_buf.size() == FLASH_SECTOR_ERASE_SIZE);
+
                     con.exit_xip();
```
lurch, line 1865: Looking at the "verify" part of the code (in the `if (settings.load.verify)` block), it seems this `con.exit_xip();` could also possibly be moved inside the `if (!skip)` block?

main.cpp
```diff
@@ -403,6 +404,7 @@ struct load_command : public cmd {
             (
                 option('n', "--no-overwrite").set(settings.load.no_overwrite) % "When writing flash data, do not overwrite an existing program in flash. If picotool cannot determine the size/presence of the program in flash, the command fails" +
                 option('N', "--no-overwrite-unsafe").set(settings.load.no_overwrite_force) % "When writing flash data, do not overwrite an existing program in flash. If picotool cannot determine the size/presence of the program in flash, the load continues anyway" +
+                option('u', "--update").set(settings.load.update) % "Skip flash writes when the data is already present" +
```
lurch, line 407: I'm not sure the help-string here is entirely clear? :shrug: 
Would something like "Skip writing flash sectors that already contain identical data" be better?

main.cpp
```diff
@@ -1859,9 +1861,18 @@ bool load_command::execute(device_map &devices) {
                     file_buf.insert(file_buf.begin(), read_range.from - aligned_range.from, 0);
                     file_buf.insert(file_buf.end(), aligned_range.to - read_range.to, 0);
                     assert(file_buf.size() == FLASH_SECTOR_ERASE_SIZE);
+
                     con.exit_xip();
-                    con.flash_erase(aligned_range.from, FLASH_SECTOR_ERASE_SIZE);
-                    raw_access.write_vector(aligned_range.from, file_buf);
+                    bool skip = false;
+                    if (settings.load.update) {
+                      vector<uint8_t> read_device_buf;
```
lurch, line 1868: Would it be better to declare this vector outside of the `for` loop? (i.e. just below `vector<uint8_t> device_buf;`)

kilograham: nah

main.cpp
rlane: Done.

main.cpp
rlane: Done.

kilograham (approved):

Issue #32, opened 2021-05-27
add a udev rule for linux systems to access the pico without sudo 


diehard67:
I created a udev rules file so I can use picotool without sudo

```
# make raspberry pi pico accessible without sudo.

SUBSYSTEM!="usb_device", ACTION!="add", GOTO="rpi2_end"
# Raspberry Pi Pico
ATTR{idVendor}=="2e8a", ATTRS{idProduct}=="0003", MODE="660", GROUP="plugdev"

LABEL="rpi2_end"
```

lurch:
I guess this is very similar to https://github.com/raspberrypi/openocd/pull/5 and https://github.com/raspberrypi/openocd/pull/27 ?
(although not identical, as it has a different `idProduct`)

diehard67:
yah sort of, this one is for a pico in BOOTSEL mode, not picoprobe, but looks like the same goal.

kilograham:
merged into develop

PR #50, opened 2022-05-10
Add udev rules for picotool

mbrunnen:
I added udev rules for devices in boot and in application mode. Probably not exhaustive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,22 +4,27 @@ You need to set PICO_SDK_PATH in the environment, or pass it to cmake.
 
 You also need to install `libusb-1.0`.
 
-Linux/Mac: use your favorite package tool. For example, on Ubuntu:
+### Linux/Mac
+
+Use your favorite package tool. For example, on Ubuntu:
 
 ```console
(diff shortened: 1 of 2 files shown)


lurch:
See also #32

mbrunnen:
> See also #32 
Ok, that's an issue, this is a pull request. We copied both from openocd, I guess.

coolnickname:
Why has this not been merged yet?

README.md
```diff
@@ -10,6 +10,12 @@ Linux/Mac: use your favorite package tool. For example, on Ubuntu:
 sudo apt install build-essential pkg-config libusb-1.0-0-dev
 ```
 
+On Linux you can add udev rules in order to use picotool without sudo:
+
+```console
+sudo cp udev/99-picotool.rules /etc/rules/99-picotool.rules
```
DavidEGrayson, line 16: I thought `/etc/udev/rules.d` was the main place to put rules like this.  I don't see /etc/rules documented [here](https://www.freedesktop.org/software/systemd/man/udev.html).

Issue #56, opened 2022-08-27
--force doesn't work on MacOS


playduck:
Executing any command with `-f` or `-F` has no effect.

picotool either responds with "No accessible RP2040 devices in BOOTSEL mode were found." or, if the device is manually put into BOOTSEL mode, it flashes the device.
Using `-f`, I'd expect the device to be rebooted. This is not the case.
Everything else regarding picotool seems to work as expected.

I'm using running on an M1 using:

- cmake version 3.24.1
- Apple clang version 13.1.6 (clang-1316.0.21.2.5) Target: arm64-apple-darwin21.6.0
- libusb-1.0, version 1.0.26
- picotool v1.1.0 (Darwin 21.6.0, AppleClang-13.1.6.13160021, Release)
- picotool pulled at `03f28122cc170535f35e5bf8fa37be024489f5eb`
- pico-sdk pulled at `2e6142b15b8a75c1227dd3edbe839193b2bf9041`

The pico is connected via a USB A to C adapter (also tested with a USB hub yielding no difference).

The cmake and make output is listed here:
```bash
➜  build git:(master) cmake ..
-- The C compiler identification is AppleClang 13.1.6.13160021
-- The CXX compiler identification is AppleClang 13.1.6.13160021
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
Using PICO_SDK_PATH from environment ('/Users/robin/Documents/code/pico/pico-sdk')
-- Found PkgConfig: /opt/homebrew/bin/pkg-config (found version "0.29.2")
-- Checking for module 'libusb-1.0'
--   Found libusb-1.0, version 1.0.26
-- Found LIBUSB: /opt/homebrew/lib/libusb-1.0.dylib
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/robin/Documents/code/pico/picotool/build
➜  build git:(master) ✗ make
[...]

PR #57, opened 2022-08-31
Add error message if --force fails

playduck:
This PR is in response to issue #56.

Using `-f` or `-F` requires the RP2040 to have enabled USB-CDC.
As this isn't immediately obvious this PR adds a short reminder if a load command with a force flag fails.
This also adds a small sentence to the `help load` text.

Running `./picotool load binary.elf -f` with no responding device now outputs:
> No accessible RP2040 devices in BOOTSEL mode were found.
> To force a device into BOOTSEL mode, make sure the RP2040 is configured to use USB-CDC (USB stdio).

The help load text now reads (note the last sentence):
> -f, --force
>             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the  command (unless the command itself is a 'reboot') the device will be rebooted back to application mode. Make sure the device is using USB-CDC (USB stdio)

I've also modified the the printing routine for the error message to use `snprintf` and `strncpy` to avoid potential string-length issues as is good practice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,12 +105,13 @@ TARGET SELECTION:
         --address <addr>
             Filter devices by USB device address
         -f, --force
-            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing
-            the command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
+            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
+            command (unless the command itself is a 'reboot') the device will be rebooted back to application mode. Make sure the device
+            is using USB CDC (USB stdio)
         -F, --force-no-reboot
-            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing
-            the command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but
-            without the RPI-RP2 drive mounted
+            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
+            command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but without
+            the RPI-RP2 drive mounted. Make sure the device is using USB CDC (USB stdio)
     To target a file
         <filename>
             The file name
@@ -213,12 +214,13 @@ OPTIONS:
         --address <addr>
             Filter devices by USB device address
         -f, --force
-            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing
-            the command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
+            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
+            command (unless the command itself is a 'reboot') the device will be rebooted back to application mode. Make sure the device
+            is using USB CDC (USB stdio)
         -F, --force-no-reboot
-            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing
-            the command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but
-            without the RPI-RP2 drive mounted
+            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
+            command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but without
+            the RPI-RP2 drive mounted. Make sure the device is using USB CDC (USB stdio)
     File to save to
         <filename>
             The file name
--- a/README.md
+++ b/README.md
@@ -216,11 +216,11 @@ OPTIONS:
         -f, --force
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be rebooted back to application mode. Make sure the device
-            is using USB CDC (USB stdio)
+            is using USB-CDC (USB stdio)
         -F, --force-no-reboot
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but without
-            the RPI-RP2 drive mounted. Make sure the device is using USB CDC (USB stdio)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -213,12 +214,13 @@ OPTIONS:
         --address <addr>
             Filter devices by USB device address
         -f, --force
-            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing
-            the command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
+            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
+            command (unless the command itself is a 'reboot') the device will be rebooted back to application mode. Make sure the device
+            is using USB CDC (USB stdio)
```
lurch, line 219: Missing hyphen in USB-CDC here?

playduck:
Added both missing dashes, must've missed them the first time around.


> It actually appears to be any command that uses the force flag

Yes, I've just been testing it with load only, but other commands behave the same:
```
./picotool info -f
To force a device into BOOTSEL mode, make sure the RP2040 is configured to use USB-CDC (USB stdio).
```

main.cpp
```diff
@@ -1564,26 +1564,31 @@ void info_guts(memory_access &raw_access) {
...
-            sprintf(buf, "accessible RP2040 devices in BOOTSEL mode were found found on bus %d.", settings.bus);
+            snprintf(bufErrorMessage, bufferLen, "accessible RP2040 devices in BOOTSEL mode were found found on bus %d.", settings.bus);
         } else {
-            sprintf(buf, "accessible RP2040 devices in BOOTSEL mode were found.");
+            snprintf(bufErrorMessage, bufferLen, "accessible RP2040 devices in BOOTSEL mode were found.");
         }
     }
+    if (settings.force) {
+        char* bufForceError = b + strnlen(b, bufferLen);
+        snprintf(bufForceError, bufferLen, "\nTo force a device into BOOTSEL mode, make sure the RP2040 is configured to use USB-CDC (USB stdio).");
```
lurch, line 1590: If you're using `snprintf` to prevent string-overflows, do you also need to pass in a smaller value than `bufferLen` if the string `b` already contains some content?

playduck: True, the same holds for the `bufErrorMessage` strings above.

The length argument should be something along the lines of `bufferLen - strnlen(b, bufferLen)`. I'm not sure if this causes an off-by-one error regarding the null terminator (e.g. should it be `bufferLen - strnlen(b, bufferLen) - 1` or not)?

lurch: > I'm not sure if this causes an off-by-one error

I dunno either :man_shrugging: I'm more of a Python programmer than a C programmer :wink:

playduck:
I subtracted the current string length from the maximum allowed length of the `snprintf` calls.
This still produces the same output but is now more robust against buffer-overflows.

main.cpp
```diff
@@ -1564,26 +1564,33 @@ void info_guts(memory_access &raw_access) {
...
-            sprintf(buf, "accessible RP2040 device in BOOTSEL mode was found at bus %d, address %d.", settings.bus, settings.address);
+            snprintf(bufErrorMessage, bufferLen - currentLength, "accessible RP2040 device in BOOTSEL mode was found at bus %d, address %d.", settings.bus, settings.address);
         } else {
-            sprintf(buf, "accessible RP2040 devices in BOOTSEL mode were found with address %d.", settings.address);
+            snprintf(bufErrorMessage, bufferLen - currentLength, "accessible RP2040 devices in BOOTSEL mode were found with address %d.", settings.address);
         }
     } else {
         if (settings.bus != -1) {
-            sprintf(buf, "accessible RP2040 devices in BOOTSEL mode were found found on bus %d.", settings.bus);
+            snprintf(bufErrorMessage, bufferLen - currentLength, "accessible RP2040 devices in BOOTSEL mode were found found on bus %d.", settings.bus);
```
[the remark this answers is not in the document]
playduck: Oh wow, my line-wrapping broke the line right between the two and I didn't notice it either.
It's fixed now.

README.md
```diff
@@ -213,12 +214,13 @@ OPTIONS:
...
         -F, --force-no-reboot
-            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing
-            the command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but
-            without the RPI-RP2 drive mounted
+            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
+            command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but without
+            the RPI-RP2 drive mounted. Make sure the device is using USB CDC (USB stdio)
```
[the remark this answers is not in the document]
DavidEGrayson: No hyphen is needed ([example](https://en.wikipedia.org/wiki/USB_communications_device_class)) and I don't think I've ever seen it written with a hyphen.  I usually write it as "USB CDC ACM".

kilograham:
it isn't strictly limited to USB stdio (though may be in practice as no one else implements the interface)

Perhaps _"Force a device not in BOOTSEL mode but running compatible code **(e.g. using USB stdio)**"_

lurch:
You probably want to change this to be based against `develop` instead of `master`? And whilst doing that, be aware of the change introduced in #72  :wink:

main.cpp
```diff
@@ -1599,9 +1600,12 @@ string missing_device_string(bool wasRetry) {
         if (settings.bus != -1) {
             snprintf(buf, buf_len, "accessible RP2040 devices in BOOTSEL mode were found found on bus %d.", settings.bus);
         } else {
-            snprintf(buf, buf_len,"accessible RP2040 devices in BOOTSEL mode were found.");
+            snprintf(buf, buf_len, "accessible RP2040 devices in BOOTSEL mode were found.");
         }
     }
+    if (settings.force) {
+        snprintf(buf, buf_len, "\nTo force a device into BOOTSEL mode, make sure the RP2040 is configured to use USB-CDC (USB stdio).");
```
lurch, line 1607: I _think_ you need to recalculate `buf` and `buf_len`, otherwise you'll just be overwriting the previous text rather than appending to it?

PR #19, opened 2021-04-01
Don't write more data than will fit in flash

lurch:

PR #73, opened 2023-02-11
Don't write more data than will fit in flash

lurch:
Updated version of #19

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -1845,12 +1845,20 @@ bool load_command::execute(device_map &devices) {
         }
     }
     auto ranges = get_coalesced_ranges(file_access);
+    bool uses_flash = false;
+    uint32_t flash_min = std::numeric_limits<uint32_t>::max();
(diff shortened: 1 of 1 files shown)


lurch:
The previous version of this PR (#19) got (accidentally?) auto-closed _again_ when the `develop` branch disappeared :confused: 

I had a quick check, and AFAICT it **looks** like none of the other PRs got auto-closed because they're all based against the `master` branch rather than the `develop` branch?  (but I didn't do a similar check for the `pico-sdk` or `pico-examples` repos, because there's lots of churn going on there and I've not been keeping up, so I wouldn't know which PRs are closed deliberately and which are closed accidentally)

kilograham:
ah yes; i need to make develop not disappear!

lurch:
@kilograham Bump, `develop` still isn't a protected branch :wink:

will-v-pi:
Thanks, this has been merged for picotool 2.0.0 - and `develop` branch is now protected too

Issue #75, opened 2023-03-05
picotool load invalid bus parsing


XDeschuyteneer:
```
picotool load ${BIN} --bus 0 --address 1 -f
```

throws:

```
ERROR: <bus> must be >= 1
```

kilograham:
merged into develop (and master)

PR #76, opened 2023-03-05
bugfix #75 : picotool load invalid bus parsing

XDeschuyteneer:


kilograham (approved):
I'm assuming 0 is a valid bus number ;-)

Issue #54, opened 2022-07-30
Feature: Target devices by it's unique ID


ggsubs:
Feature request: Target devices by their unique ID, so that I can work with multiple devices connected at the same time, while allowing the USB port connections to change.

The unique id is available on the SDK (pico_get_unique_board_id), the same string is exposed on the device descriptor.

ggsubs:
Example: slusb -v output shows iSerial value:

```
Bus 002 Device 004: ID 2e8a:000a  
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.00
  bDeviceClass          239 Miscellaneous Device
  bDeviceSubClass         2 
  bDeviceProtocol         1 Interface Association
  bMaxPacketSize0        64
  idVendor           0x2e8a 
  idProduct          0x000a 
  bcdDevice            1.00
  iManufacturer           1 Raspberry Pi
  iProduct                2 Pico
  iSerial                 3 E66038B713249237
  bNumConfigurations      1
```

neilenns:
Yes, this would be excellent. I opened similar issues a while back (https://github.com/raspberrypi/picotool/issues/48) and despite how I closed that issue I actually never solved my problem. I couldn't find any way to programmatically reference an RP2040 device and flash it using picotool.

lurch:
I guess this ties up closely with #12 (since the ID returned by `pico_get_unique_board_id` **is** the flash-chip's unique ID.

Ferdi265:
I have created a very experimental branch [here](https://github.com/ferdi265/picotool/tree/rewrite-main) that provides a `--serial` option for selecting devices, as well as `--multiple` for e.g., rebooting multiple devices into BOOTSEL mode at once.

The branch is pretty experimental, but it simplifies the workflow for working with multiple picos by a lot.

For example:
```
picotool load server.uf2 --serial $SERVER_SERIAL -f
picotool load client.uf2 --serial $CLIENT_SERIAL -f
```
will correctly reboot two specific picos into BOOTSEL mode, flash them, and reboot them back into application mode.
I'm still cleaning up the branch, but I hope I will have something quite usable in a week or two.

will-v-pi:
This has been added with picotool 2.0.0 as the `--ser` option

PR #83, opened 2023-05-24
Add stdio_usb reset interface detection for custom VendorID / ProductID setups

tomas-pecserke:
It would be really nice to use `picotool` with projects using custom VID/PID as well.

With the new Pico SDK 1.5.0 having improved compatibility of stdio_usb with TinyUSB it's easy to implement stdio_usb reset feature with custom USB setup.

With this patch I am able to reboot the Pico running a TinyUSB stack into BOOTSEL mode and upload new firmware automatically.

That been said I still have several issues I'd like  some feedback on:
1) With non-standard VID / PID supported, it would be helpful to add device filters based on (at least) these attributes:
    - Vendor ID - PR #85
    - Product ID - PR #85
    - Serial ID - PR #84 
2) User might want to have the CLI option to enable/disable this feature. :heavy_check_mark: *Implemented*
3) UDEV rule based on VENDOR interface and stdio_usb reset interface subclass and protocol might be nice, but the attributes required seem to be very distribution dependent. This prevents definition of reliable rule to apply in all situations. As this is advanced use case anyway, additional step required to allow specific user's device is acceptable. :x: *Will not implement*

**Example use case:**
I would like to create automated upload cmake target for a Pico project:
```cmake
# Custom command to upload firmware to Raspberry Pi Pico
add_custom_target(upload_firmware
    COMMAND ${PICOTOOL_EXECUTABLE} load -x ${TARGET_NAME}.uf2 --detect-reset-interface -f
    DEPENDS ${TARGET_NAME}
    COMMENT "Uploading firmware to Raspberry Pi Pico"
)
```

I also implemented a tiny library that simplyfies the implementation of custom TinyUSB reset interface - 
[pico_tusb_reset_interface](https://github.com/tomas-pecserke/pico_tusb_reset_interface).

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,14 +70,14 @@ PICOTOOL:
     Tool for interacting with a RP2040 device in BOOTSEL mode, or with a RP2040 binary
 
 SYNOPSIS:
-    picotool info [-b] [-p] [-d] [-l] [-a] [--bus <bus>] [--address <addr>] [-f] [-F]
+    picotool info [-b] [-p] [-d] [-l] [-a] [--bus <bus>] [--address <addr>] [-i] [-f] [-F]
     picotool info [-b] [-p] [-d] [-l] [-a] <filename> [-t <type>]
-    picotool load [-n] [-N] [-u] [-v] [-x] <filename> [-t <type>] [-o <offset>] [--bus <bus>] [--address <addr>] [-f] [-F]
-    picotool save [-p] [--bus <bus>] [--address <addr>] [-f] [-F] <filename> [-t <type>]
-    picotool save -a [--bus <bus>] [--address <addr>] [-f] [-F] <filename> [-t <type>]
-    picotool save -r <from> <to> [--bus <bus>] [--address <addr>] [-f] [-F] <filename> [-t <type>]
-    picotool verify [--bus <bus>] [--address <addr>] [-f] [-F] <filename> [-t <type>] [-r <from> <to>] [-o <offset>]
-    picotool reboot [-a] [-u] [--bus <bus>] [--address <addr>] [-f] [-F]
+    picotool load [-n] [-N] [-u] [-v] [-x] <filename> [-t <type>] [-o <offset>] [--bus <bus>] [--address <addr>] [-i] [-f] [-F]
+    picotool save [-p] [--bus <bus>] [--address <addr>] [-i] [-f] [-F] <filename> [-t <type>]
+    picotool save -a [--bus <bus>] [--address <addr>] [-i] [-f] [-F] <filename> [-t <type>]
+    picotool save -r <from> <to> [--bus <bus>] [--address <addr>] [-i] [-f] [-F] <filename> [-t <type>]
+    picotool verify [--bus <bus>] [--address <addr>] [-i] [-f] [-F] <filename> [-t <type>] [-r <from> <to>] [-o <offset>]
+    picotool reboot [-a] [-u] [--bus <bus>] [--address <addr>] [-i] [-f] [-F]
     picotool version [-s]
     picotool help [<cmd>]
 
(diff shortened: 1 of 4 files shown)


lurch:
Just to "keep the dots connected", #54 is also requesting filtering by Serial ID.

README.md
```diff
@@ -133,6 +133,8 @@ TARGET SELECTION:
             Filter devices by USB bus number
         --address <addr>
             Filter devices by USB device address
+        -i, --enable-reset-interface
+            Enable detection of stdio_usb-compatible reset interface for custom USB devices
```
lurch, line 137: Only a small niggle, but IMHO `--enable-reset-interface` sounds like it's doing something "active", and different to "Enable detection of ...", if you see what I mean? :thinking: 
Perhaps `--reset-interface` is more "passive" and more in line with the other `--bus`, `--address` and `--force` options?

Also, I wonder if it _might_ be worth adding more of an explanation (elsewhere in this README) of what "detection of stdio_usb-compatible reset interface" means? :man_shrugging:

tomas-pecserke: These options all do something active, I want to make it clear this adds support for the reset interface. How about `--support-reset-interface` or `--allow-reset-interface`? I feel like just `--reset-interface` is not clear enough.

lurch: > I feel like just `--reset-interface` is not clear enough.

True, I hadn't considered that `--reset-interface` could itself also be misinterpreted as an "active" thing :laughing: 

> How about `--support-reset-interface` or `--allow-reset-interface`?

Yeah they sound like decent options. The commit-comment says "Add CLI option to enable detection of devices with reset interface", so perhaps `--detect-reset-interfaces` is another alternative? :man_shrugging:

main.cpp
```diff
@@ -2276,19 +2283,25 @@ int main(int argc, char **argv) {
 #if defined(_WIN32)
                             printer(dr_vidpid_stdio_usb,
                                     " appears to be a RP2040 device with a USB serial connection, not in BOOTSEL mode. You can force reboot into BOOTSEL mode via 'picotool reboot -f -u' first.");
+                            printer(dr_reset_interface,
+                                    " appears to be a RP2040 device with a USB reset interface, not in BOOTSEL mode. You can force reboot into BOOTSEL mode via 'picotool reboot -f -u' first.");
```
lurch, line 2287: Should the example picotool command-line here also include the `-i` flag?

tomas-pecserke: Right. The users will only see this message, if they already are using '-i' option. Do we want to promote the possibility of using the option similar to how we promote use of `-f` option?

tomas-pecserke:
@lurch if you are interested, pls have a look at a companion library implementing the reset interface for custom TinyUSB stack:
https://github.com/tomas-pecserke/pico_tusb_reset_interface

lurch:
I'm not the `picotool` maintainer (that's @kilograham ), I just try to leave helpful comments for the little problems that I happen to spot :slightly_smiling_face:

lurch:
> @lurch if you are interested, pls have a look at a companion library implementing the reset interface for custom TinyUSB stack: https://github.com/tomas-pecserke/pico_tusb_reset_interface

Hmmm, not sure if it's wise to say "[picotool](https://github.com/raspberrypi/picotool) supports custom devices with compatible reset interface with use of --detect-reset-interface or -i option." if this picotool PR hasn't even been merged yet? Probably best to just add a link to this PR for the time being? :thinking:

will-v-pi:
Closing this as #177 has been merged which allows the rebooting of a device with custom `--vid` and `--pid` - this means you can reboot a device into bootsel mode and then load firmware with:
```
picotool reboot -f -u --vid 0xabcd --pid 0x1234
picotool load myfile.elf
```

PR #85, opened 2023-06-03
Add VID and PID device target filters

tomas-pecserke:
This PR improves usability of #83 and depends on #84.

It adds `--vid <vid>` and `--pid <pid>` device target filter options to device selection accepting hex id representation.
The Vendor ID and Product ID is then compared with identifiers specified by `idVendor` and `idProduct` of device configuration option respectively.

This solution does NOT break backward compatibility - functionality and outputs do not change if the flag is not specified.

Usage:
```shell
# TinyUSB examples use Cafe:0200
picotool info --pid cafe --vid 0200
```

**❗Attention:** Help needed with clipp options for these filters.
Currently I have all the options as optional and despite that i get this error when used without any options:
```shell
$ picotool info

ERROR: missing required argument

SYNOPSIS:
    picotool info [-b] [-p] [-d] [-l] [-a] [--bus <bus>] [--address <addr>] [--vid <vid>] [--pid <pid>] [--serial <serial>] [-f] [-F]
    picotool info [-b] [-p] [-d] [-l] [-a] <filename> [-t <type>]

Use "picotool help info" for more info
```
If you have any idea why, please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,14 +70,14 @@ PICOTOOL:
     Tool for interacting with a RP2040 device in BOOTSEL mode, or with a RP2040 binary
 
 SYNOPSIS:
-    picotool info [-b] [-p] [-d] [-l] [-a] [--bus <bus>] [--address <addr>] [--serial <serial>] [-f] [-F]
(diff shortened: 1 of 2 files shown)


lurch:
This should probably be targeting the `develop` branch?

will-v-pi:
These have been added for picotool 2.0.0

PR #86, opened 2023-06-03
Add serial command that prints the flash unique id

Ferdi265:
This PR adds a command to display the flash unique id from boards currently in BOOTSEL mode. (This PR replaces #80)

Since the bootrom RPC interface does not have a way to directly issue flash commands such as  `CMD_RUID`, the only way I found to fix this is to copy the `fash_get_unique_id` function to RAM and use the `EXEC` command to read out the unique id.

the code blob uploaded to the pico was compiled using the following C and ASM code:

```c
#include "hardware/regs/io_qspi.h"
#include "hardware/structs/ioqspi.h"
#include "hardware/structs/ssi.h"
#include "hardware/flash.h"

asm(
    ".macro static_assert value, msg\n"
        ".if !(\\value)\n"
        ".err \\msg\n"
        ".endif\n"
    ".endm\n"

    ".set FLASH_RUID_CMD, 0x4b\n"
    ".set FLASH_RUID_DUMMY_BYTES, 4\n"
    ".set FLASH_RUID_DATA_BYTES, 8\n"
    ".set FLASH_RUID_TOTAL_BYTES, (1 + FLASH_RUID_DUMMY_BYTES + FLASH_RUID_DATA_BYTES)\n"
);

void flash_do_cmd(const uint8_t * txbuf, uint8_t *rxbuf, size_t count);
void __attribute__((naked)) flash_get_unique_id_raw(void) {
    asm(
        ".Lflash_get_unique_id_raw:\n"
            "adr r0, .Ltxbuf\n"
            "adr r1, .Lrxbuf\n"
            "ldr r2, .Lbuflen\n"
            "b flash_do_cmd\n"
        ".Lbuflen:\n"
            ".word FLASH_RUID_TOTAL_BYTES\n"
        ".Ltxbuf:\n"
            ".byte FLASH_RUID_CMD\n"
            ".zero (FLASH_RUID_TOTAL_BYTES - 1)\n"
            ".zero (16 - FLASH_RUID_TOTAL_BYTES)\n"
        ".Lrxbuf:\n"
            ".zero FLASH_RUID_TOTAL_BYTES\n"
            ".zero (16 - FLASH_RUID_TOTAL_BYTES)\n"
        "static_assert ((.Lrxbuf - flash_get_unique_id_raw) == 28), \"rxbuf offset incorrect\"\n"
    );
}

static inline void __attribute__((always_inline)) flash_cs_force(_Bool high) {
    uint32_t field_val = high ?
        IO_QSPI_GPIO_QSPI_SS_CTRL_OUTOVER_VALUE_HIGH :
        IO_QSPI_GPIO_QSPI_SS_CTRL_OUTOVER_VALUE_LOW;
    hw_write_masked(&ioqspi_hw->io[1].ctrl,
        field_val << IO_QSPI_GPIO_QSPI_SS_CTRL_OUTOVER_LSB,
        IO_QSPI_GPIO_QSPI_SS_CTRL_OUTOVER_BITS
    );
}

void flash_do_cmd(const uint8_t * txbuf, uint8_t *rxbuf, size_t count) {
    flash_cs_force(0);
    size_t tx_remaining = count;
    size_t rx_remaining = count;
    // We may be interrupted -- don't want FIFO to overflow if we're distracted.
    const size_t max_in_flight = 16 - 2;
    while (tx_remaining || rx_remaining) {
        uint32_t flags = ssi_hw->sr;
        bool can_put = !!(flags & SSI_SR_TFNF_BITS);
        bool can_get = !!(flags & SSI_SR_RFNE_BITS);
        if (can_put && tx_remaining && rx_remaining - tx_remaining < max_in_flight) {
            ssi_hw->dr0 = *txbuf++;
            --tx_remaining;
        }
        if (can_get && rx_remaining) {
            *rxbuf++ = (uint8_t)ssi_hw->dr0;
            --rx_remaining;
        }
    }
    flash_cs_force(1);
}
```

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -346,6 +346,23 @@ struct info_command : public cmd {
     }
 };
 
+struct serial_command : public cmd {
+    serial_command() : cmd("serial") {}
+    bool execute(device_map& devices) override;
+
+    group get_cli() override {
+        return (
+            (
+                device_selection % "To target one or more connected RP2040 device(s) in BOOTSEL mode (the default)"
+            ).major_group("TARGET SELECTION").min(0).doc_non_optional(true)
+         );
+    }
+
+    string get_doc() const override {
+         return "Display the serial number of the target device in BOOTSEL mode";
+    }
+};
+
 struct verify_command : public cmd {
     verify_command() : cmd("verify") {}
     bool execute(device_map &devices) override;
@@ -495,6 +512,7 @@ struct reboot_command : public cmd {
 
 vector<std::shared_ptr<cmd>> commands {
         std::shared_ptr<cmd>(new info_command()),
+        std::shared_ptr<cmd>(new serial_command()),
         std::shared_ptr<cmd>(new load_command()),
         std::shared_ptr<cmd>(new save_command()),
         std::shared_ptr<cmd>(new verify_command()),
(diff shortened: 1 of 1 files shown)


will-v-pi:
Thanks for this - it has been merged for picotool 2.0.0 as part of the `picotool info` device section, and also gets used when picking an RP2040 device by serial number

Ferdi265:
@will-v-pi for completeness, here is a gist with a CMake project that generates the binary blob from this PR exactly:

https://gist.github.com/Ferdi265/2b6a3ffb808d5b1a3d2465342259cb6c

The code was compiled with pico-sdk 2.0.0 and arm-none-eabi-gcc 14.1.0.

PR #87, opened 2023-08-19
Use tags to determine file permissions instead of groups

Ktrod:
`99-picotool.rules` uses the `plugdev` group which Arch doesn't ship with by default. 

This change removes the `plugdev` group in favor of adding the `uaccess` tag and changes the priority from `99` to `71`. 

Since this is built into `systemd` [by default ](https://github.com/systemd/systemd/blob/main/rules.d/70-uaccess.rules.in) it's distro-agnostic


will-v-pi:
We opted to use #99 in lieu of this, which adds `uaccess` without removing `plugdev`, and leaves the priority at 99

PR #97, opened 2024-06-05
set default terminal size if ioctl fails

josch:
This is necessary to run picotool with tools that consume its --help output like help2man. Otherwise, terminal width will be set to bogus values like zero on i386 or 52446 on x86_64.


will-v-pi:
Thanks - this has been merged for picotool 2.0.0

PR #99, opened 2024-06-06
udev/99-picotool.rules: use TAG+="uaccess"

josch:
allow the currently logged in user to access the device


will-v-pi:
Thanks - this has been merged with picotool 2.0.0

PR #101, opened 2024-06-18
Add Bazel build

armandomontanez:
Adds a Bazel build for Picotool. The Bazel build comes with a few perks:
1. You don't need to set `PICO_SDK_PATH`. It's still possible to locally override with `--override_module=pico-sdk=/path/to/pico-sdk`.
2. Libusb is built from source. The version to use can be controlled in dependent projects by using `rules_libusb` version constraints.
3. Windows works out of the box with just MSVC.


BUILD.bazel
```diff
@@ -0,0 +1,44 @@
...
+        "clipp/clipp.h",
+        "elf.h",
+        "main.cpp",
+        "picoboot_connection/picoboot_connection.c",
+        "picoboot_connection/picoboot_connection.h",
+        "picoboot_connection/picoboot_connection_cxx.cpp",
+        "picoboot_connection/picoboot_connection_cxx.h",
+    ],
+    copts = select({
+        "@platforms//os:windows": [],
```
tpudlik, line 16: You need to add `bazel_dep(name = "platforms", version="0.0.7")` to the `MODULE.bazel` because of this `@platforms` reference.

kilograham (approved):

PR #102, opened 2024-07-15
Improve including LIBUSB on Windows

micropicostack:
Add additional checks to determine which library to use - MinGW or VS and if 32bit or 64bit library. Allow to use PkgConfig on Windows as well if it is installed (in PATH).


micropicostack:
Initial commit had an issue with the `PkgConfig` check. The idea was to check if `pkg-config` tool is available then use that (linux and windows) if not revert to manual path hints settings. Moved the package check before checking `PKG_CONFIG_FOUND`. 

On windows it is possible to install `pkg-config` with `winget`
```
winget install bloodrock.pkg-config-lite
```
Once this is done, set the `PKG_CONFIG_PATH`from where the `pkg-config` will find the `.pc` files. 
```
setx PKG_CONFIG_PATH C:\Work\PkgConfigs
```

After this is done the, it is possible to build `picotool` with the pkg config

Example libusb.pc file
```
prefix=C:/Work/libusb-1.0.27
exec_prefix=${prefix}
libdir=${exec_prefix}/MinGW64/static
includedir=${prefix}/include

Name: libusb-1.0
Description: Lib USB
Version: 1.0.27
Requires:
Conflicts:
Libs:
Cflags:
```

I hope this helps others! 😄

will-v-pi:
Thanks for this - it has been merged for picotool 2.0.0

PR #103, opened 2024-08-07
CMakeLists.txt: also link to additional package-config libraries

ghost:
LibUSB might specify additional libraries to link to via package-config (i.e. libatomic where this isn't a builtin [1]). These are correctly discovered by the call to pkg_check_modules and put into the PC_LIBUSB_LIBRARIES variable but then never added to the actual linking step.

Fix that by additionally adding them behind the libusb library discovered by the find_library call().

[1] https://github.com/libusb/libusb/blob/v1.0.27/configure.ac#L184

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,7 +58,8 @@ else()
             pico_platform_headers
             pico_usb_reset_interface_headers
             picoboot_connection_cxx
-            ${LIBUSB_LIBRARIES})
+            ${LIBUSB_LIBRARIES}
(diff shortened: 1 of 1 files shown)


will-v-pi (changes requested):
I think it would be better to make this change in cmake/FindLIBUSB.cmake and append `PC_LIBUSB_LIBRARIES` to the `LIBUSB_LIBRARIES` variable, rather than using `PC_LIBUSB_LIBRARIES` in the main CMakeLists.txt

Also, could you target this at the develop branch instead of master

PR #109, opened 2024-08-10
Clarify picotool builds when libusb is not present.

DarwinJS:
Add clarification of picotool builds that are missing all the usb interface commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Use your favorite package tool to install dependencies. For example, on Ubuntu:
 sudo apt install build-essential pkg-config libusb-1.0-0-dev cmake
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,6 +16,8 @@ Use your favorite package tool to install dependencies. For example, on Ubuntu:
 sudo apt install build-essential pkg-config libusb-1.0-0-dev cmake
 ```
 
+> If libusb-1.0-0-dev is not present, picotool still builds, but it automatically leave out all options that deal with managing a pico via usb (load, save, verify, reboot). This type of build can be recognized because these commands also do not appear in the help command.
```
lurch, line 19: "not present" -> "not installed" ?
"it automatically leave out" -> "it omits" ?
"usb" -> "USB" ?

PR #116, opened 2024-08-19
Support multiple tries when using -f/F

will-v-pi:
This is necessary for WSL, or other cases where it takes more time to detect the device after reboot, without increasing the delay for everyone

@kilograham We could possibly also decrease the normal timeout from 1200ms, I'm not sure how long it normally takes to reboot?

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -75,6 +75,8 @@ static __forceinline int __builtin_ctz(unsigned x) {
 #define _CRT_SECURE_NO_WARNINGS
 #endif
 
+#define MAX_REBOOT_TRIES 5
+
 #define OTP_PAGE_COUNT 64
 #define OTP_PAGE_ROWS  64
 #define OTP_ROW_COUNT (OTP_PAGE_COUNT * OTP_PAGE_ROWS)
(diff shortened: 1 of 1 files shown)


main.cpp
```diff
@@ -7448,49 +7450,57 @@ int main(int argc, char **argv) {
...
+                                    fos << "\nbut:\n\n";
+                                    had_note = true;
+                                }
+                                for (auto d : devices[r]) {
+                                    fos << bus_device_string(std::get<1>(d)) << description << "\n";
+                                }
+                            };
+    #if defined(__linux__) || defined(__APPLE__)
+                            printer(dr_vidpid_bootrom_cant_connect,
+                                    " appears to be a RP2040 device in BOOTSEL mode, but picotool was unable to connect. Maybe try 'sudo' or check your permissions.");
```
lurch, line 7477: Are these error messages definitely only applicable to RP2040, or do they apply to RP2350 too?

will-v-pi: They're for both, but that's outside the scope of this PR (I think there are a few other places where it'd need fixing too) - this was just an indentation change

PR #121, opened 2024-08-26
Fix loading into PSRAM

will-v-pi:
Skip the flash size checks when loading into PSRAM, as they are only designed for the CS0 window

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -4146,7 +4146,8 @@ bool load_guts(picoboot::connection con, iostream_memory_access &file_access) {
         uint32_t flash_start_offset = flash_min - FLASH_START;
         uint32_t size_guess = guess_flash_size(raw_access);
         if (size_guess > 0) {
-            if ((flash_start_offset + flash_data_size) > size_guess) {
+            // Skip check when targeting PSRAM, which is anything above 0x11000000
+            if ((flash_start_offset + flash_data_size) > size_guess && flash_start_offset < FLASH_END_RP2040) {
                 if (flash_start_offset) {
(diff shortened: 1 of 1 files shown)


main.cpp
```diff
@@ -4146,7 +4146,8 @@ bool load_guts(picoboot::connection con, iostream_memory_access &file_access) {
         uint32_t flash_start_offset = flash_min - FLASH_START;
         uint32_t size_guess = guess_flash_size(raw_access);
         if (size_guess > 0) {
-            if ((flash_start_offset + flash_data_size) > size_guess) {
+            // Skip check when targeting PSRAM, which is anything above 0x11000000
+            if ((flash_start_offset + flash_data_size) > size_guess && flash_start_offset < FLASH_END_RP2040) {
```
lurch, line 4150: Only a minor style thing, but I think I'd be inclined to swap the conditions around to `if (flash_start_offset < FLASH_END_RP2040 && (flash_start_offset + flash_data_size) > size_guess)` to better indicate that "if `flash_start_offset` is beyond `FLASH_END_RP2040` then don't even bother _trying_ to see if it fits" ? :wink:

will-v-pi: Yeah, that makes more sense

kilograham (approved):

lurch:
Does this mean that picotool would "happily" try loading a binary into PSRAM, that's bigger than the available PSRAM?

will-v-pi:
Yes - but I'm not sure if there's a way to detect/guess the size of the available PSRAM, like we currently do for flash?

PR #122, opened 2024-08-26
Add support for multi-family UF2s to picotool info

will-v-pi:
This adds support for UF2s with multiple families to picotool info, and also throws a warning that it is only using the first family for other commands.

For example output from the sample hello_world.uf2
```
$ picotool info hello_world.uf2 
File hello_world.uf2 family ID 'rp2040':

Program Information
 name:                  hello_universal
 features:              UART stdin / stdout
                        USB stdin / stdout
 binary start:          0x10000000
 binary end:            0x10005dd0
 target chip:           RP2350
 image type:            ARM Secure
 rolling window delta:  0x00006000
 vector table:          0x10000000

-------------------------------------------
File hello_world.uf2 family ID 'absolute':

Program Information
 name:                  hello_universal
 features:              UART stdin / stdout
                        USB stdin / stdout
 binary start:          0x10000000
 binary end:            0x10005dd0
 target chip:           RP2350
 image type:            ARM Secure
 rolling window delta:  0x00006000
 vector table:          0x10000000
```

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -2626,10 +2626,11 @@ void build_rmap_elf(std::shared_ptr<std::iostream>file, range_map<size_t>& rmap)
     }
 }
 
-void build_rmap_uf2(std::shared_ptr<std::iostream>file, range_map<size_t>& rmap) {
+uint32_t build_rmap_uf2(std::shared_ptr<std::iostream>file, range_map<size_t>& rmap, uint32_t family_id=0) {
     file->seekg(0, ios::beg);
     uf2_block block;
     unsigned int pos = 0;
+    uint32_t next_family_id = 0;
     do {
         file->read((char*)&block, sizeof(uf2_block));
         if (file->fail()) {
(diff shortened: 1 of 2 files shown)


lurch:
I have zero knowledge about this "universal" UF2 stuff, but why does the first part of your example above say both "family ID 'rp2040'" and "target chip: RP2350" ? :thinking:

main.cpp
```diff
@@ -3718,16 +3741,34 @@ bool config_command::execute(device_map &devices) {
...
+                    string dashes;
+                    std::generate_n(std::back_inserter(dashes), s.str().length() + 1, [] { return '-'; });
+                    fos << "\n" << dashes << "\n";
+                }
+                fos << s.str() << "\n\n";
+                auto tmp_access = get_file_memory_access(0, false, &next_id);
+                info_guts(tmp_access, nullptr);
+            }
+        } else {
+            fos << "File " << settings.filenames[0] << ":\n\n";
```
lurch, line 3769: Not directly-related to this change, but is it also worth displaying the family id for chips that _don't_ have a `next_id` ?

will-v-pi:
> I have zero knowledge about this "universal" UF2 stuff, but why does the first part of your example above say both "family ID 'rp2040'" and "target chip: RP2350" ? 🤔

The actual "universal binary" (which is just duplicated twice in the "universal UF2" with different family IDs so it gets loaded onto rp2040 or rp2350) has metadata blocks for RP2040, RP2350-Arm and RP2350-RiscV - currently `picotool info` just displays information about the "best" block found, which will be the RP2350-Arm block. We should probably add support for displaying info about all metadata blocks, rather than just the "best" block it chooses, but that's a separate issue

main.cpp
will-v-pi: Added, so it now displays family id for all UF2 files, not just multi-family UF2 files

lurch:
BTW, on the current `develop` branch of picotool:
```console
$ grep "[fF]amily id" main.cpp | wc -l
13
$ grep "[fF]amily ID" main.cpp | wc -l
3
```

For consistency, should those occurrences of "family id" be changed to "family ID" ?

kilograham (approved):

PR #124, opened 2024-08-26
Small refactoring of the hex_string functions

lurch:


will-v-pi:
Thanks for this - I think it's actually better to just remove the `HEX_string` function entirely, as it's only used once. If you're happy with that then I'll merge this?

lurch:
I'm happy! :slightly_smiling_face:

PR #125, opened 2024-08-28
Add verify option to picotool save

### LINKED ISSUE #113: would be nice for "save" to support verify like load
I am forever doing this after!

will-v-pi:
Adds an option to verify files after saving. Stops verification at the end of the saved region, to avoid issues where a UF2 is padded with 0s, but flash may have been erased then written without padding so ends with `0xff`s

Fixes #113


kilograham (approved):
This mixes in some help output from another change (family id), but that can be worked out when submitting other things I guess

PR #128, opened 2024-08-30
README.md: Improve the MSYS2 and installation instructions

DavidEGrayson:
I originally contributed those MSYS2 instructions, and they used Make.  Now I realize it's better to let MSYS2 use its default CMake generator, which is Ninja.  (You can see the commands got a lot simpler now that we are using Ninja.)

The installation instructions below that point that assumed that Make was being used, because it said `make install`, but we don't want to use Make on MSYS2, so I changed it to `cmake --install .`.

I also noticed that the installation instructions could be improved in other ways.  There was a typo (picootol), some misplaced commas, shorthand (ELF->UF2), and passive voice.  It did not mention the option of setting picotool_DIR as an environment variable (which is very handy), and it was setting picotool_DIR equal to the path to the executable instead of the directory, which seems wrong.  I did test setting the picotool_DIR environment variable equal to a directory, and it works as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,32 +66,23 @@ No need to download libusb separately or set `LIBUSB_ROOT`.
 pacman -S $MINGW_PACKAGE_PREFIX-{toolchain,cmake,libusb}
 mkdir build
 cd build
-MSYS2_ARG_CONV_EXCL=- cmake .. -G"MSYS Makefiles" -DCMAKE_INSTALL_PREFIX=$MINGW_PREFIX
-make
-make install DESTDIR=/  # optional
+cmake .. -DCMAKE_INSTALL_PREFIX=$MINGW_PREFIX
+cmake --build .
 ```
 
 ## Usage by the Raspberry Pi Pico SDK
 
-The Raspberry Pi Pico SDK ([pico-sdk](https://github.com/raspberrypi/pico-sdk)) version 2.0.0 and above, uses `picotool` to do the ELF->UF2 conversion previously handled by the `elf2uf2` tool in the SDK. `picootol` is also used by the SDK for hashing and/or signing binaries.
+The Raspberry Pi Pico SDK ([pico-sdk](https://github.com/raspberrypi/pico-sdk)) version 2.0.0 and above uses `picotool` to do the ELF-to-UF2 conversion previously handled by the `elf2uf2` tool in the SDK. The SDK also uses `picotool` to hash and sign binaries.
 
-Whilst the SDK can download picotool on its own per project, if you have multiple projects or build configurations, it is preferable to install a single copy of `picotool` locally.
+Whilst the SDK can download picotool on its own per project, if you have multiple projects or build configurations, it is preferable to install a single copy of `picotool` locally.  This can be done most simply with `make install` or `cmake --install .`; the SDK will use this installed version by default.
 
-This can be done most simply with `make install`; the SDK will use this installed version by default.
-
-Alternatively you can install in a custom path via:
+Alternatively, you can install to a custom path via:
 
 ```
 cmake -DCMAKE_INSTALL_PREFIX=$MY_INSTALL_DIR -DPICOTOOL_FLAT_INSTALL=1 ..
 ```
 
-In order for the SDK to find `picotool` in this custom path, you  will need to set the `picotool_DIR` variable in your project, either by passing to `-Dpicotool_DIR=$MY_INSTALL_DIR/picotool` to your SDK `cmake` command, or by adding
-
-```CMake
-set(picotool_DIR $MY_INSTALL_DIR/picotool)
-```
-
-to your CMakeLists.txt file.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,32 +66,23 @@ No need to download libusb separately or set `LIBUSB_ROOT`.
 pacman -S $MINGW_PACKAGE_PREFIX-{toolchain,cmake,libusb}
```
will-v-pi, line 66: Do you need to add a ninja installation to this pacman command too, or is that installed by default (or by one of these packages already)?

README.md
```diff
@@ -66,32 +66,23 @@ No need to download libusb separately or set `LIBUSB_ROOT`.
...
 
 ## Usage by the Raspberry Pi Pico SDK
 
-The Raspberry Pi Pico SDK ([pico-sdk](https://github.com/raspberrypi/pico-sdk)) version 2.0.0 and above, uses `picotool` to do the ELF->UF2 conversion previously handled by the `elf2uf2` tool in the SDK. `picootol` is also used by the SDK for hashing and/or signing binaries.
-
-Whilst the SDK can download picotool on its own per project, if you have multiple projects or build configurations, it is preferable to install a single copy of `picotool` locally.
+The Raspberry Pi Pico SDK ([pico-sdk](https://github.com/raspberrypi/pico-sdk)) version 2.0.0 and above uses `picotool` to do the ELF-to-UF2 conversion previously handled by the `elf2uf2` tool in the SDK. The SDK also uses `picotool` to hash and sign binaries.
 
-This can be done most simply with `make install`; the SDK will use this installed version by default.
+Whilst the SDK can download picotool on its own per project, if you have multiple projects or build configurations, it is preferable to install a single copy of `picotool` locally.  This can be done most simply with `cmake --install .`; the SDK will use this installed version by default.
```
will-v-pi, line 77: I'd prefer to still make reference to `make install` here, and just add `cmake --install .` as another option
```suggestion
Whilst the SDK can download picotool on its own per project, if you have multiple projects or build configurations, it is preferable to install a single copy of `picotool` locally.  This can be done most simply with `make install` or `cmake --install .`; the SDK will use this installed version by default.
```

README.md
```diff
@@ -66,32 +66,23 @@ No need to download libusb separately or set `LIBUSB_ROOT`.
...
 ```
 
-In order for the SDK to find `picotool` in this custom path, you  will need to set the `picotool_DIR` variable in your project, either by passing to `-Dpicotool_DIR=$MY_INSTALL_DIR/picotool` to your SDK `cmake` command, or by adding
-
-```CMake
-set(picotool_DIR $MY_INSTALL_DIR/picotool)
-```
-
-to your CMakeLists.txt file.
+In order for the SDK to find `picotool` in this custom path, you will need to set the `picotool_DIR` variable in your project, either by setting the `picotool_DIR` environment variable, by passing `-Dpicotool_DIR=$MY_INSTALL_DIR` to your `cmake` command, or by adding `set(picotool_DIR $MY_INSTALL_DIR)` to your CMakeLists.txt file.
```
will-v-pi, line 85: The `$MY_INSTALL_DIR/picotool` path is actually a directory still - the executable is at `$MY_INSTALL_DIR/picotool/picotool`. For the environment variable it works to set `picotool_DIR` to `$MY_INSTALL_DIR` or `$MY_INSTALL_DIR/picotool`, but for setting the CMake variable it must be set to `$MY_INSTALL_DIR/picotool` else it doesn't work. So these lines still need to say `$MY_INSTALL_DIR/picotool`

will-v-pi (changes requested):
Thanks for this PR, I just had a few comments but happy to merge once those are addressed

README.md
DavidEGrayson: Nope, ninja is one of the [dependencies of CMake in MSYS2](https://packages.msys2.org/package/mingw-w64-x86_64-cmake) so it will always be installed if CMake is installed.

README.md
DavidEGrayson: Oh, sorry for not looking at this carefully the first time.  I just force-pushed to fix this.

README.md
DavidEGrayson: OK, sure, I just force-pushed to address this.

DavidEGrayson:
Thanks for reviewing this so quickly!  I force-pushed a new commit that should address everything.

will-v-pi (approved):
Thanks, these changes look good - I'll merge this

lurch:
Oops, looks like this was merged onto `master` instead of `develop`. (or was that deliberate @will-v-pi ?)

will-v-pi:
Not deliberate, and I've updated the branch protection rules to reduce the risk of this in future

lurch:
Might be worth cherry-picking that commit onto `develop`, and reverting it on `master` ?

will-v-pi:
I've merged `master` into `develop` to pick up this change - I think it's fine to leave it on `master` without reverting as there is a `2.0.0` tag which is used by the SDK

PR #131, opened 2024-09-02
Add github actions compile test

will-v-pi:
Add compilation test across variations of OS, generator, and with/without Libusb, MbedTLS, and pre-compiled xip-ram-perms and flash-id executables

### DIFF
--- a/.github/workflows/choco_packages.config
+++ b/.github/workflows/choco_packages.config
@@ -4,6 +4,6 @@
   <package id="cmake" version="3.25.2" installArguments="ADD_CMAKE_TO_PATH=System" />
   <package id="mingw" version="12.2.0" />
   <package id="ninja" version="1.11.1" />
-	<package id="7zip.install" version="24.8.0" />
+  <package id="7zip.install" version="24.8.0" />
   <package id="wget" version="1.21.4" />
 </packages>
--- a/.github/workflows/choco_packages.config
+++ b/.github/workflows/choco_packages.config
@@ -4,6 +4,4 @@
   <package id="cmake" version="3.25.2" installArguments="ADD_CMAKE_TO_PATH=System" />
   <package id="mingw" version="12.2.0" />
   <package id="ninja" version="1.11.1" />
-  <package id="7zip.install" version="24.8.0" />
-  <package id="wget" version="1.21.4" />
 </packages>
--- a/.github/workflows/choco_packages.config
+++ b/.github/workflows/choco_packages.config
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="gcc-arm-embedded" version="10.2.1" />
+  <package id="cmake" version="3.25.2" installArguments="ADD_CMAKE_TO_PATH=System" />
+  <package id="mingw" version="12.2.0" />
+  <package id="ninja" version="1.11.1" />
+	<package id="7zip.install" version="24.8.0" />
+</packages>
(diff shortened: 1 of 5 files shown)


.github/workflows/choco_packages.config
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="gcc-arm-embedded" version="10.2.1" />
+  <package id="cmake" version="3.25.2" installArguments="ADD_CMAKE_TO_PATH=System" />
+  <package id="mingw" version="12.2.0" />
+  <package id="ninja" version="1.11.1" />
+	<package id="7zip.install" version="24.8.0" />
```
lurch, line 7: Inconsistent indent :stuck_out_tongue_winking_eye:

will-v-pi: Oops - thanks

ndabas:
If you wanted to be a bit more minimal about the Windows build, the Windows runner [already has a bunch of stuff installed](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md) so you don't need to install the `choco` packages:
- CMake is preinstalled.
- MinGW is likely not needed? It seems you're just using `bash` at some point but PowerShell could be used too. If you're installing `ninja` anyway you wont need `make` either.
- 7zip is preinstalled.
- You could use `curl` instead of `wget` because `curl` is included with Windows anyway.

will-v-pi:
> * CMake is preinstalled.
> * MinGW is likely not needed? It seems you're just using `bash` at some point but PowerShell could be used too. If you're installing `ninja` anyway you wont need `make` either.
>   
>   * msys2 is preinstalled (but not added to PATH) if you really need it.

The reason for installing CMake and MinGW from chocolatey is so we have known compilers, and a know installation process that we can suggest to users, rather than just using the versions built into GitHub actions runners

> * 7zip is preinstalled.
> * You could use `curl` instead of `wget` because `curl` is included with Windows anyway.

Thanks for these suggestions, I will switch the action to use curl and the pre-installed 7zip

lurch:
Looks like the pre-installed 7zip is causing the Windows tests to fail? :confused:

ndabas:
> Looks like the pre-installed 7zip is causing the Windows tests to fail? 😕

I guess you need to call `curl` as `curl.exe` on Windows because `curl` is an alias to `Invoke-WebRequest` or something silly in PowerShell by default.

will-v-pi:
> I guess you need to call `curl` as `curl.exe` on Windows because `curl` is an alias to `Invoke-WebRequest` or something silly in PowerShell by default. I mean I think it's not the 7zip that's failing, it's the curl before that.

I agree it looks like the curl that's failing, but it looks like it is invoking `curl.exe`, as `Invoke-WebRequest` doesn't give the `% Total % Received` etc printout.

will-v-pi:
> Yes, you're right, it is invoking curl, I guess you just need a flag to allow redirects with `curl`.

Thanks, that's it - I'll use `curl -L`

ndabas:
> The reason for installing CMake and MinGW from chocolatey is so we have known compilers, and a know installation process that we can suggest to users, rather than just using the versions built into GitHub actions runners

On that topic, unrelated to this particular PR, it might make sense to write a small PowerShell script that installs these dependencies the same way the VS Code extension does? Basically just use the manifest files that are used there, download and add to PATH for command line builds. I see the SDK, examples, and now this (picotool) have copies of this choco-based install, which are not exactly the same; and one of the major differences from the VS Code ext is the ARM GCC specified for choco is much older (even pico-setup-windows had a newer build.)

Issue #139, opened 2024-09-09
recompile ./xip_ram_perms/xip_ram_perms.elf requires PICOTOOL_NO_LIBUSB


josch:
Hi,

for the Debian packaging of picotool 2.0, I'd like to rebuild the file `./xip_ram_perms/xip_ram_perms.elf` but when reading `CMakeLists.txt` it seems that doing so is only possible if one also sets `PICOTOOL_NO_LIBUSB=1`. But I'd like to build picotool with libusb support. Am I misreading something? Can I somehow build picotool with libusb *and* rebuild `./xip_ram_perms/xip_ram_perms.elf`?

Also, why is `./xip_ram_perms/xip_ram_perms.elf` tracked in git?

lurch:
Already fixed by https://github.com/raspberrypi/picotool/pull/131/commits/c4b8630d499592330c563a6b6a2a782562d3c009 which is part of https://github.com/raspberrypi/picotool/pull/131

will-v-pi:
I think you're misreading the code - it's the other way round, so it only compiles `xip_ram_perms.elf` if `PICOTOOL_NO_LIBUSB=0` (or isn't defined), and `USE_PRECOMPILED=false` must also be set. I've merged #131 now, which fixes it downloading and building picotool inside the `xip_ram_perms` build

The reason for including it in the git repository is that users who don't wish to compile from source (for example if they don't have a working pico-sdk build) should still be able to build picotool. This same reasoning applies to the `picoboot_flash_id/flash_id.bin` build

will-v-pi:
You can see [this job](https://github.com/raspberrypi/picotool/actions/runs/10771070045/job/29865802869) for an example compilation with libusb and compiling these binaries from source

josch:
Thank you, it seems that I followed a red-herring and mis-read the if-condition in CMakeLists.txt. I'm currently configuring picotool with `-DPICO_SDK_PATH=/usr/src/pico-sdk -DUSE_PRECOMPILED=0` and it seems to work because it wants to build pico-sdk now... which fails becaus:

```
CMake Warning at /usr/src/pico-sdk/tools/Findpicotool.cmake:28 (message):
  No installed picotool with version 2.0.0 found - building from source

  It is recommended to build and install picotool separately, or to set
  PICOTOOL_FETCH_FROM_GIT_PATH to a common directory for all your SDK
  projects
```

[...]

PR #140, opened 2024-09-09
xip_ram_perms/CMakeLists.txt: unset environment variables CFLAGS, CXX…

josch:
…FLAGS and LDFLAGS

If the user set these environment variables to influence the picotool build, unset them here so that they do not influence the pico-sdk build. This is especially required for flags that are not supported by arm-none-eabi compilers.

As suggested by @will-v-pi in https://github.com/raspberrypi/picotool/issues/139#issuecomment-2338077730

### DIFF
--- a/picoboot_flash_id/CMakeLists.txt
+++ b/picoboot_flash_id/CMakeLists.txt
@@ -6,6 +6,14 @@ if (NOT USE_PRECOMPILED)
     # default build type
     set(CMAKE_BUILD_TYPE "MinSizeRel" CACHE STRING "build type")
 
+    # If the user set these environment variables to influence the picotool
+    # build, unset them here so that they do not influence the pico-sdk
+    # build. This is especially required for flags that are not supported
+    # by arm-none-eabi compilers.
+    unset(ENV{CFLAGS})
+    unset(ENV{CXXFLAGS})
+    unset(ENV{LDFLAGS})
+
     include(${PICO_SDK_PATH}/external/pico_sdk_import.cmake)
     project(flash_id C CXX ASM)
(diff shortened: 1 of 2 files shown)


will-v-pi:
Could you target this at the develop branch rather than master?

Also, I think the same changes should probably be added the the `picoboot_flash_id/CMakeLists.txt` file as well, as that also uses a pico-sdk build

lurch:
What if the user wants to set different CFLAGS for the `picotool` build and the `xip_ram_perms` build? Would it be worth supporting e.g. a `PICO_SDK_CFLAGS` setting, which would be used as `CFLAGS` when building `xip_ram_perms` and `picoboot_flash_id` ? :shrug:  (Apologies in advance if this is a stupid question!)

will-v-pi:
> What if the user wants to set different CFLAGS for the `picotool` build and the `xip_ram_perms` build? Would it be worth supporting e.g. a `PICO_SDK_CFLAGS` setting, which would be used as `CFLAGS` when building `xip_ram_perms` and `picoboot_flash_id` ? 🤷 (Apologies in advance if this is a stupid question!)

I can’t really think of a use case for that, as the SDK sets all the relevant `CFLAGS` required to build correctly? Especially for a minimal binary like this, I’m not sure what extra `CFLAGS` might be required. So I think we can merge this as is, and revisit if anyone comes up with a need for a `PICO_SDK_CFLAGS` setting

lurch:
> I can’t really think of a use case for that, as the SDK sets all the relevant `CFLAGS` required to build correctly?

Fair enough, I guess I was just playing devil's advocate :laughing:

PR #146, opened 2024-09-19
Add note about pico-sdk-tools to README.md

### LINKED ISSUE #218: Document issue: sign command does not exist in `otp` / `load` section
Inside README.md `otp` / `load` section there is an example 
```
picotool sign hello_world.elf hello_world.signed.elf private.pem otp.json
......
```

But actually `sign` command cannot be found

```
ERROR: Unknown command: sign
```

will-v-pi:
Add note about pre-built picotool binaries in the pico-sdk-tools repository

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -8,7 +8,7 @@ git submodule update --init lib/mbedtls
 
 You also need to install `libusb-1.0` if you want to use the USB functionality.
 
-> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands also do not appear in the help command. The build output message 'libUSB is not found - no USB support will be built' also appears in the build logs.
+> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands won't appear in the output of the help command. The build output message 'libUSB is not found - no USB support will be built' also appears in the build logs.
 
 ### Linux / macOS
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -8,6 +8,8 @@ git submodule update --init lib/mbedtls
 
 You also need to install `libusb-1.0` if you want to use the USB functionality.
 
+> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands also do not appear in the help command. The build output message 'libUSB is not found - no USB support will be built' also appears in the build logs.
+
 ### Linux / macOS
 
 Use your favorite package tool to install dependencies. For example, on Ubuntu:
@@ -15,8 +17,6 @@ Use your favorite package tool to install dependencies. For example, on Ubuntu:
 ```console
 sudo apt install build-essential pkg-config libusb-1.0-0-dev cmake
 ```
-
-> If libusb-1.0-0-dev is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands also do not appear in the help command. The build output message 'libUSB is not found - no USB support will be built' also appears in the build logs.
 
 Then simply build like a normal CMake project:
 
@@ -39,7 +39,7 @@ sudo cp udev/99-picotool.rules /etc/udev/rules.d/
 
 Download libUSB from here https://libusb.info/
 
-set LIBUSB_ROOT environment variable to the install directory.
+Set LIBUSB_ROOT environment variable to the install directory.
 ```console
 mkdir build
 cd build
@@ -51,7 +51,7 @@ nmake
 
 Download libUSB from here https://libusb.info/
 
-set LIBUSB_ROOT environment variable to the install directory.
+Set LIBUSB_ROOT environment variable to the install directory.
 
 ```console
 mkdir build
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,0 +1,99 @@
+## Building
+
+You need to set PICO_SDK_PATH in the environment, or pass it to cmake with `-DPICO_SDK_PATH=/path/to/pico-sdk`. To use features such as signing or hashing, you will need to make sure the mbedtls submodule in the SDK is checked out - this can be done by running this from your SDK directory.
+
+```console
+git submodule update --init lib/mbedtls
+```
+
+You also need to install `libusb-1.0` if you want to use the USB functionality.
+
+### Linux / macOS
+
+Use your favorite package tool to install dependencies. For example, on Ubuntu:
+
+```console
+sudo apt install build-essential pkg-config libusb-1.0-0-dev cmake
+```
+
+> If libusb-1.0-0-dev is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands also do not appear in the help command. The build output message 'libUSB is not found - no USB support will be built' also appears in the build logs.
+
+Then simply build like a normal CMake project:
+
+```console
+mkdir build
+cd build
+cmake ..
+make
+```
+
+On Linux you can add udev rules in order to run picotool without sudo:
+
+```console
+sudo cp udev/99-picotool.rules /etc/udev/rules.d/
+```
+
+### Windows
+
+##### For Windows without MinGW
+
+Download libUSB from here https://libusb.info/
+
+set LIBUSB_ROOT environment variable to the install directory.
+```console
+mkdir build
+cd build
+cmake -G "NMake Makefiles" ..
+nmake
+```
+
+##### For Windows with MinGW in WSL
+
+Download libUSB from here https://libusb.info/
+
+set LIBUSB_ROOT environment variable to the install directory.
+
+```console
+mkdir build
+cd build
+cmake ..
+make
+```
+
+##### For Windows with MinGW in MSYS2:
+
+No need to download libusb separately or set `LIBUSB_ROOT`.
+
+```console
+pacman -S $MINGW_PACKAGE_PREFIX-{toolchain,cmake,libusb}
+mkdir build
+cd build
+cmake .. -DCMAKE_INSTALL_PREFIX=$MINGW_PREFIX
+cmake --build .
+```
+
+## Installing (so the Pico SDK can find it)
+
+The Raspberry Pi Pico SDK ([pico-sdk](https://github.com/raspberrypi/pico-sdk)) version 2.0.0 and above uses `picotool` to do the ELF-to-UF2 conversion previously handled by the `elf2uf2` tool in the SDK. The SDK also uses `picotool` to hash and sign binaries.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,5 +1,7 @@
 ## Building
 
+> If you don't need to build picotool yourself, you can find pre-built binaries for Windows, macOS, and Linux in the [pico-sdk-tools](https://github.com/raspberrypi/pico-sdk-tools/releases) repository
```
lurch, line 3: "If you don't need..." sounds a bit strange - why would anybody **need** to build picotool themselves? :wink: 

Perhaps this could be shortened to something like "Pre-built binaries of picotool for Windows, macOS, and Linux can be downloaded from the [pico-sdk-tools](https://github.com/raspberrypi/pico-sdk-tools/releases) repository. But if you want to build it yourself from source..."

Hmm, it's obviously a slightly larger change, but given that we _are_ now offering pre-built binaries, perhaps the "Building" section should come _after_ the "Overview" section? Or perhaps be moved to an entirely separate `BUILDING.md` file?

will-v-pi: I think that keeping the Building and Usage by the SDK sections at the top is useful, as the Overview is very long and therefore people might miss the build instructions, which would probably create more issues.

> "If you don't need..." sounds a bit strange - why would anybody **need** to build picotool themselves? 😉

You might not trust binaries you've downloaded from the internet, or you might want to install picotool system-wide - both of these would require building it yourself, although maybe "If you don't want..." would be better phrasing? I think it's still the main use case for people who arrive at this repository, and building it from source should work fine across all platforms - it works fine on Windows for me

lurch: Fair enough. But perhaps it's worth having a brief "Introduction" or "Description" at the top of `README.md`, rather than diving straight into "Building" on the very first line? :shrug:

kilograham: I agree with @lurch - possibly move the building into a separate BUILDING.md and link from the other

Note also I think the linked getting-started-doc has less info than is in the README,

Open to suggestion about where to put the "Use by SDK" stuff - probably in the BUILDING.md but note in the main README link that the BUILDING.md (should it be INSTALL.md) also explains how to build/install to prevent it needing to be installed on the fly - perhaps even just keep the section on use by the SDK, but just have it explain that SDK uses it, and you can install locally to not need it build every time, and then link to the BUILDING.md

will-v-pi: I've restructured it, so the main README just has a small section linking to BUILDING.md, and includes the details about pico-sdk-tools. I've also added some hopefully helpful notes to prevent people doing the wrong thing (eg installing it incorrectly by just copying it).

BUILDING.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Linux / macOS
+
+Use your favorite package tool to install dependencies. For example, on Ubuntu:
+
+```console
+sudo apt install build-essential pkg-config libusb-1.0-0-dev cmake
+```
+
+> If libusb-1.0-0-dev is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands also do not appear in the help command. The build output message 'libUSB is not found - no USB support will be built' also appears in the build logs.
```
lurch: Should this be moved out of the "Linux / macOS" section, and into the previous section where it  says "You also need to install `libusb-1.0` if you want to use the USB functionality." ?

BUILDING.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+You also need to install `libusb-1.0` if you want to use the USB functionality.
+
+### Linux / macOS
+
+Use your favorite package tool to install dependencies. For example, on Ubuntu:
+
+```console
+sudo apt install build-essential pkg-config libusb-1.0-0-dev cmake
```
lurch, line 18: Is there a simple `brew` command to install the needed dependencies on macOS?

BUILDING.md
```diff
@@ -0,0 +1,99 @@
...
+sudo cp udev/99-picotool.rules /etc/udev/rules.d/
+```
+
+### Windows
+
+##### For Windows without MinGW
+
+Download libUSB from here https://libusb.info/
+
+set LIBUSB_ROOT environment variable to the install directory.
```
lurch: nit: capital 'S' on "set" ? (also on line 54_

README.md
```diff
@@ -155,6 +57,12 @@ Use "picotool help <cmd>" for more info
 
 Note commands that aren't acting on files require a device in BOOTSEL mode to be connected.
 
+## Building & Installing
+
+If you don't want to build picotool yourself, you can find pre-built executables for Windows, macOS, and Linux in the [pico-sdk-tools](https://github.com/raspberrypi/pico-sdk-tools/releases) repository. Assuming you've unzipped that directory to `$PICOTOOL_LOC` (with the actual picotool executable at `$PICOTOOL_LOC/picotool/picotool`), you can point the Pico SDK at this binary by setting the `picotool_DIR` environment variable to `$PICOTOOL_LOC/picotool`, or by passing `-Dpicotool_DIR=$PICOTOOL_LOC/picotool` to your `cmake` command or setting it in your `CMakeLists.txt` file.
```
lurch: Maybe this would be slightly less confusing if you used `$SDKTOOLS_LOC` instead of `$PICOTOOL_LOC` ? (assuming of course that I haven't misunderstood what's going on! :wink: )

lurch: "unzipped that directory" doesn't quite sound right?

will-v-pi: Yeah, I wasn't really sure how to phrase it - I want to say that if you download it to Downloads/xxx.zip, extract that to Downloads/xxx, then the picotool binary is at Downloads/xxx/picotool/picotool - the crucial point is that it's not at Downloads/xxx/picotool

will-v-pi: The releases contain `pico-sdk-tools` releases (currently just contains `pioasm`) and `picotool` releases as separate files, so I think `SDKTOOLS_LOC` might be confusing?

BUILDING.md
will-v-pi: Possibly just `brew install pkgconf libusb cmake`? But you also need to install the Xcode command line tools, or some alternative compiler that CMake can find

README.md
lurch: Perhaps `$EXTRACT_LOCATION` ? My concern with `$PICOTOOL_LOC/picotool/picotool` is that people might find it confusing with the word "picotool" appearing 3 times?

lurch: I think that depending on the "unzip" program being used and/or which unzip options are chosen, you might end up with `picotool-2.1.1-x64-win\picotool\picotool.exe` or you might end up with just `picotool\picotool.exe` ?
Either way, I think "extracted that archive" might sound better than "unzipped that directory"?

BUILDING.md
lurch: Hmmm, in that case perhaps the "macOS" instructions should be broken out into their own section?

will-v-pi: Maybe a separate PR to improve (and test) the MacOS instructions once this one has been merged in? This was just copied directly from the existing README

BUILDING.md
```diff
@@ -8,6 +8,8 @@ git submodule update --init lib/mbedtls
 
 You also need to install `libusb-1.0` if you want to use the USB functionality.
 
+> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands also do not appear in the help command. The build output message 'libUSB is not found - no USB support will be built' also appears in the build logs.
```
lurch: I think "because these commands also do not appear in the help command." should perhaps be reworded to "because these commands won't appear in the output of the help command." ?

README.md
```diff
@@ -155,6 +57,12 @@ Use "picotool help <cmd>" for more info
 
 Note commands that aren't acting on files require a device in BOOTSEL mode to be connected.
 
+## Building & Installing
+
+If you don't want to build picotool yourself, you can find pre-built executables for Windows, macOS, and Linux in the [pico-sdk-tools](https://github.com/raspberrypi/pico-sdk-tools/releases) repository. Assuming you've extracted that archive to `$EXTRACT_LOCATION` (with the actual picotool executable at `$EXTRACT_LOCATION/picotool/picotool`), you can point the Pico SDK at this binary by setting the `picotool_DIR` environment variable to `$EXTRACT_LOCATION/picotool`, or by passing `-Dpicotool_DIR=$EXTRACT_LOCATION/picotool` to your `cmake` command or setting it in your `CMakeLists.txt` file.
```
lurch: Reading this again, would using `<extract_location>` to signify a placeholder-directory work better than using `$EXTRACT_LOCATION`? Because otherwise some users might think that they need to set an `$EXTRACT_LOCATION` environment variable? :thinking:

lurch (approved):
One small question, but otherwise LGTM

README.md
will-v-pi: Yeah, that's clearer
```suggestion
If you don't want to build picotool yourself, you can find pre-built executables for Windows, macOS, and Linux in the [pico-sdk-tools](https://github.com/raspberrypi/pico-sdk-tools/releases) repository. Assuming you've extracted that archive to `<extract_location>` (with the actual picotool executable at `<extract_location>/picotool/picotool`), you can point the Pico SDK at this binary by setting the `picotool_DIR` environment variable to `<extract_location>/picotool`, or by passing `-Dpicotool_DIR=<extract_location>/picotool` to your `cmake` command or setting it in your `CMakeLists.txt` file.
```

lurch (approved):

kilograham (approved):
nice

BUILDING.md
```diff
@@ -0,0 +1,99 @@
+## Building
+
+You need to set PICO_SDK_PATH in the environment, or pass it to cmake with `-DPICO_SDK_PATH=/path/to/pico-sdk`. To use features such as signing or hashing, you will need to make sure the mbedtls submodule in the SDK is checked out - this can be done by running this from your SDK directory.
+
+```console
+git submodule update --init lib/mbedtls
```
lurch, line 6: Hmmm, in light of other recent changes, does this now also need to update mbedtls's submodules too?

will-v-pi: It won't once #226 is merged, so I'll leave this as is

lurch: So now that **this** PR has been merged, does that imply that #226 ought to be rebased, and make its own changes to `README.md`? Or am I just getting confused by all the overlapping PRs again? :rofl:

will-v-pi: #226 removes the need for mbedtls submodules, so won't need the readme update

lurch: Right, so does #226 also need to update `BUILDING.md` to remove the `git submodule update` command, or am I still getting my wires crossed?

will-v-pi: You still need the SDK's mbedtls submodule (pico-sdk/lib/mbedtls), but you don't need Mbedtls's submodules (pico-sdk/lib/mbedtls/framework etc) - so the command is still needed, but a new command to do it recursively instead is not needed, so no changes need to be made to the readme.

lurch: Aha, thanks! It's the "you don't need Mbedtls's submodules (pico-sdk/lib/mbedtls/framework etc)" bit that I didn't realise, and is why I was getting confused :rofl:   But in that case, why is / was https://github.com/raspberrypi/pico-examples/pull/635 needed?

will-v-pi: That works around the "problem" that #226 fixes, but probably isn't necessary if #226 will be merged soon, although won't do any harm either

will-v-pi: Basically if you consider needing Mbedtls submodules to be the bug, then #226 is the bug-fix, and raspberrypi/pico-examples#635, https://github.com/raspberrypi/pico-sdk/pull/2416, 8be60e9eeab96857c84d71f0565cad0e84654c77 are all workarounds to get CI working

PR #154, opened 2024-10-15
Use RP-series when referring to RP2040/RP2350

will-v-pi:
In most cases picotool still referred to RP2040, with one instance of RP2040/RP2350. Standardise all these to RP2xxx when it could be referring to either RP2040 or RP2350.

@kilograham or @lurch Any opinions on RP2040/RP2350 vs RP2xxx? I went with RP2xxx as it's shorter, but RP2040/RP2350 would work fine too if preferred.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -151,8 +151,16 @@ std::array<std::array<string, 48>, 12> pin_functions_rp2350{{
 std::map<uint32_t, otp_reg> otp_regs;
 
 #if HAS_LIBUSB
-auto bus_device_string = [](struct libusb_device *device) {
-    return string("Device at bus ") + std::to_string(libusb_get_bus_number(device)) + ", address " + std::to_string(libusb_get_device_address(device));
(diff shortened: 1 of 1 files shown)


kilograham:
Yeah I was kinda hoping to actually use the correct name but that is maybe overkill; I assume there are some places we don't know the answer which would still need a generic term. I forget what term we use in the documentation... sorry havent read this change yet

will-v-pi:
CC @nathan-contino to check what term we use in the documentation - it might be RP2-series or something?

nathan-contino:
In the docs, we've landed on "RP-series microcontroller" for this exact concept; you can likely shorten that to "RP-series chips" or "RP-series silicon" etc based on the specific situation. No need to lock ourselves into RP2 :-)

kilograham:
yeah, i think RP-series device is good (for when we don't actually know)

kilograham (approved):

PR #159, opened 2024-10-25
Fix coprodis to not drop the last instruction and the rest of the file

Daft-Freak:
Don't break from the loop early so that we copy the remaining lines (and also actually output the last translated instruction too).

Disassembly is more useful when most of it isn't missing! :laughing:


kilograham:
ah thank you; i have been wondering what that was!

PR #160, opened 2024-10-25
some tiny improvements and way more comments

SimonWaldherr:


will-v-pi:
Could you target this at the develop branch instead of master

PR #162, opened 2024-10-29
some tiny improvements and way more comments

SimonWaldherr:
pull request changed to "develop"-branch as requested by @will-v-pi (https://github.com/raspberrypi/picotool/pull/160)

### DIFF
--- a/elf2uf2/elf2uf2.cpp
+++ b/elf2uf2/elf2uf2.cpp
@@ -11,49 +11,93 @@
 #include <cstring>
 #include <cstdarg>
 #include <algorithm>
-#include <cstring>
 #include <memory>
 
 #include "elf2uf2.h"
 #include "errors.h"
 
+// Size of a flash sector in bytes
 #define FLASH_SECTOR_ERASE_SIZE 4096u
 
+// Define the page size for UF2 (256 Bytes)
+// These definitions are already present in the header file and should not be redefined here.
+// #define LOG2_PAGE_SIZE 8u
+// #define UF2_PAGE_SIZE (1u << LOG2_PAGE_SIZE)
+
+// Global variable to control verbose mode
 static bool verbose;
 
+/**
+ * @brief Terminates the program with a read error.
+ */
 static void fail_read_error() {
     fail(ERROR_READ_FAILED, "Failed to read input file");
 }
 
+/**
+ * @brief Terminates the program with a write error.
+ */
 static void fail_write_error() {
     fail(ERROR_WRITE_FAILED, "Failed to write output file");
 }
 
+/**
+ * @struct page_fragment
+ * @brief Represents a fragment of a page with file offset and size.
+ */
 struct page_fragment {
-    page_fragment(uint32_t file_offset, uint32_t page_offset, uint32_t bytes) : file_offset(file_offset), page_offset(page_offset), bytes(bytes) {}
-    uint32_t file_offset;
-    uint32_t page_offset;
-    uint32_t bytes;
+    /**
+     * @brief Constructor for page_fragment.
+     * @param file_offset Offset in the input file.
+     * @param page_offset Offset within the page.
+     * @param bytes Number of bytes in the fragment.
+     */
+    page_fragment(uint32_t file_offset, uint32_t page_offset, uint32_t bytes)
+        : file_offset(file_offset), page_offset(page_offset), bytes(bytes) {}
+
+    uint32_t file_offset;  ///< Offset in the input file.
+    uint32_t page_offset;  ///< Offset within the page.
+    uint32_t bytes;        ///< Number of bytes in the fragment.
 };
 
+/**
+ * @brief Checks if an address is within the valid ranges.
+ * 
+ * @param valid_ranges Valid address ranges.
+ * @param addr Physical address to check.
+ * @param vaddr Virtual address.
+ * @param size Size of the area to check.
+ * @param uninitialized Flag indicating if the area is uninitialized.
+ * @param ar Reference to an address_range that will be set on success.
+ * @return 0 on success, error code otherwise.
+ */
 int check_address_range(const address_ranges& valid_ranges, uint32_t addr, uint32_t vaddr, uint32_t size, bool uninitialized, address_range &ar) {
     for(const auto& range : valid_ranges) {
         if (range.from <= addr && range.to >= addr + size) {
             if (range.type == address_range::type::NO_CONTENTS && !uninitialized) {
-                fail(ERROR_INCOMPATIBLE, "ELF contains memory contents for uninitialized memory at %p", addr);
+                fail(ERROR_INCOMPATIBLE, "ELF contains memory contents for uninitialized memory at 0x%08x", addr);
             }
             ar = range;
             if (verbose) {
-                printf("%s segment %08x->%08x (%08x->%08x)\n", uninitialized ? "Uninitialized" : "Mapped", addr,
-                   addr + size, vaddr, vaddr+size);
+                printf("%s segment 0x%08x->0x%08x (0x%08x->0x%08x)\n", 
+                       uninitialized ? "Uninitialized" : "Mapped", 
+                       addr, addr + size, vaddr, vaddr + size);
             }
             return 0;
         }
     }
-    fail(ERROR_INCOMPATIBLE, "Memory segment %08x->%08x is outside of valid address range for device", addr, addr+size);
+    fail(ERROR_INCOMPATIBLE, "Memory segment 0x%08x->0x%08x is outside of valid address range for device", addr, addr + size);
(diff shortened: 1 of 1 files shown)


elf2uf2/elf2uf2.cpp
```diff
@@ -118,13 +174,26 @@ int realize_page(std::shared_ptr<std::iostream> in, const std::vector<page_fragm
...
+    // TODO: Check actual address within page
     return true;
 }
 
+/**
+ * @brief Generates an absolute UF2 block.
+ * 
+ * @param abs_block_loc Target address for the absolute block.
+ * @return Generated UF2 block.
+ */
```
will-v-pi, line 196: It would probably be good to mention Errata E10 in this comment and the one below - it's not just any absolute block, it's an absolute block required to work around Errata E10 in the datasheet

elf2uf2/elf2uf2.cpp
```diff
@@ -204,10 +304,12 @@ int bin2uf2(std::shared_ptr<std::iostream> in, std::shared_ptr<std::iostream> ou
     if (size <= 0) {
         fail_read_error();
     }
+    in->seekg(0, in->beg); // Reset to beginning
```
will-v-pi, line 307: This isn't actually required, as the function that reads the file (`realize_page`) seeks to the appropriate place in the file before reading

elf2uf2/elf2uf2.cpp
```diff
@@ -231,80 +333,100 @@ int bin2uf2(std::shared_ptr<std::iostream> in, std::shared_ptr<std::iostream> ou
...
+    uint32_t thumb_bit = (eh.common.machine == EM_ARM) ? 0x1u : 0x0u;
+
     if (ram_style) {
         uint32_t expected_ep_main_ram = UINT32_MAX;
         uint32_t expected_ep_xip_sram = UINT32_MAX;
+
+        // Determine expected entry points
         for(auto& page_entry : pages) {
-            if ( ((page_entry.first >= SRAM_START) && (page_entry.first < ram_range[0].to)) && (page_entry.first < expected_ep_main_ram) ) {
+            if (((page_entry.first >= SRAM_START) && (page_entry.first < ram_range[0].to)) && (page_entry.first < expected_ep_main_ram)) {
```
will-v-pi, line 392: These lines could instead have the second set of brackets on the left removed as it's just a 3-way AND, to prevent the `(((` which is hard to read
```suggestion
            if ((page_entry.first >= SRAM_START) && (page_entry.first < ram_range[0].to) && (page_entry.first < expected_ep_main_ram)) {
```

elf2uf2/elf2uf2.cpp
```diff
@@ -11,49 +11,93 @@
...
 #include "elf2uf2.h"
 #include "errors.h"
 
+// Size of a flash sector in bytes
 #define FLASH_SECTOR_ERASE_SIZE 4096u
 
+// Define the page size for UF2 (256 Bytes)
+// These definitions are already present in the header file and should not be redefined here.
+// #define LOG2_PAGE_SIZE 8u
+// #define UF2_PAGE_SIZE (1u << LOG2_PAGE_SIZE)
```
will-v-pi, line 25: These lines can be removed - the definitions had already been removed from this cpp file

will-v-pi (commented):
Thanks, these changes look good - there are a couple things I'd want changed, but probably worth a review from @lurch too?

elf2uf2/elf2uf2.cpp
```diff
@@ -11,49 +11,93 @@
...
+    uint32_t page_offset;  ///< Offset within the page.
+    uint32_t bytes;        ///< Number of bytes in the fragment.
 };
 
+/**
+ * @brief Checks if an address is within the valid ranges.
+ * 
+ * @param valid_ranges Valid address ranges.
+ * @param addr Physical address to check.
+ * @param vaddr Virtual address.
```
lurch, line 68: Could add " (only used in verbose mode)" ?

elf2uf2/elf2uf2.cpp
```diff
@@ -11,49 +11,93 @@
...
+ * @param size Size of the area to check.
+ * @param uninitialized Flag indicating if the area is uninitialized.
+ * @param ar Reference to an address_range that will be set on success.
```
lurch, line 71: "Reference to the valid_range that the provided (addr + size) fits inside" ?

elf2uf2/elf2uf2.cpp
```diff
@@ -11,49 +11,93 @@
...
             return 0;
         }
     }
-    fail(ERROR_INCOMPATIBLE, "Memory segment %08x->%08x is outside of valid address range for device", addr, addr+size);
+    fail(ERROR_INCOMPATIBLE, "Memory segment 0x%08x->0x%08x is outside of valid address range for device", addr, addr + size);
     return ERROR_INCOMPATIBLE;
 }
 
+/**
+ * @brief Checks ELF-32 program header entries and populates the page fragments map.
```
lurch, line 94: @will-v-pi Is this hyphenation of "ELF-32" correct? :man_shrugging:

elf2uf2/elf2uf2.cpp
```diff
@@ -11,49 +11,93 @@
...
+    page_fragment(uint32_t file_offset, uint32_t page_offset, uint32_t bytes)
+        : file_offset(file_offset), page_offset(page_offset), bytes(bytes) {}
+
+    uint32_t file_offset;  ///< Offset in the input file.
+    uint32_t page_offset;  ///< Offset within the page.
+    uint32_t bytes;        ///< Number of bytes in the fragment.
 };
 
+/**
+ * @brief Checks if an address is within the valid ranges.
```
lurch, line 64: It checks both `addr` and `addr + size`, so perhaps this should say "Checks if a region is within the valid ranges." ?

elf2uf2/elf2uf2.cpp
```diff
@@ -193,9 +282,20 @@ int pages2uf2(std::map<uint32_t, std::vector<page_fragment>>& pages, std::shared
...
 }
 
+/**
+ * @brief Converts a binary file to UF2 format.
+ * 
+ * @param in Input stream for the binary file.
+ * @param out Output stream for UF2 data.
+ * @param address Start address in the target memory.
+ * @param family_id Family ID for UF2.
+ * @param abs_block_loc Optional absolute block location.
```
lurch, line 292: If this is optional, what special value indicates "don't use this"?

elf2uf2/elf2uf2.cpp
```diff
@@ -231,80 +333,100 @@ int bin2uf2(std::shared_ptr<std::iostream> in, std::shared_ptr<std::iostream> ou
...
 }
 
+/**
+ * @brief Converts an ELF file to UF2 format.
+ * 
+ * @param in Input stream for the ELF file.
+ * @param out Output stream for UF2 data.
+ * @param family_id Family ID for UF2.
+ * @param package_addr Optional packaging address.
+ * @param abs_block_loc Optional absolute block location.
```
lurch, line 343: Same questions about "Optional"

elf2uf2/elf2uf2.cpp
will-v-pi: I think it should be un-hyphenated, so ELF32

elf2uf2/elf2uf2.cpp
```diff
@@ -118,13 +169,26 @@ int realize_page(std::shared_ptr<std::iostream> in, const std::vector<page_fragm
...
 static bool is_address_mapped(const std::map<uint32_t, std::vector<page_fragment>>& pages, uint32_t addr) {
     uint32_t page = addr & ~(UF2_PAGE_SIZE - 1);
     if (!pages.count(page)) return false;
-    // todo check actual address within page
+    // TODO: Check actual address within page
     return true;
 }
 
+/**
+ * @brief Generates an absolute UF2 block. The absolute block is required to work around Errata E10 in the RP2040 datasheet.
```
lurch, line 187: I suspect you meant the RP2350 datasheet here, not RP2040?
https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf
https://datasheets.raspberrypi.com/rp2350/rp2350-datasheet.pdf

elf2uf2/elf2uf2.cpp
will-v-pi: A 0 means don't add this - any non-zero value means do use this

lurch: Could you add this to the doxygen please @SimonWaldherr ?

SimonWaldherr:
So, what's the status? Would you like any other changes?

will-v-pi:
There are a couple of unresolved comments from @lurch still - but once those are resolved this can be merged

PR #163, opened 2024-11-01
Add an entry point when signing Arm images

will-v-pi:
Reads the entry point and stack pointer from the vector table, and adds them to the image_def as a new entry_point item. This is only done for signing Arm executable images.

This assumes that the vector table is at 0x10000000 unless a vector_table metadata item is present, which the SDK adds for no_flash binaries, so it should work with all SDK binaries.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -4607,6 +4607,29 @@ void sign_guts_elf(elf_file* elf, private_t private_key, public_t public_key) {
         new_block.items.push_back(version);
     }
 
+    // Add entry point when signing Arm images
+    std::shared_ptr<image_type_item> image_type = new_block.get_item<image_type_item>();
+    if (settings.seal.sign && image_type != nullptr && image_type->image_type() == type_exe && image_type->cpu() == cpu_arm) {
+        std::shared_ptr<entry_point_item> entry_point = new_block.get_item<entry_point_item>();
+        if (entry_point == nullptr) {
+            std::shared_ptr<vector_table_item> vtor = new_block.get_item<vector_table_item>();
+            uint32_t vtor_loc = 0x10000000;
+            if (vtor != nullptr) {
+                vtor_loc = vtor->addr;
+            }
+            auto segment = elf->segment_from_physical_address(vtor_loc);
+            auto content = elf->content(*segment);
+            auto offset = vtor_loc - segment->physical_address();
+            uint32_t ep;
+            memcpy(&ep, content.data() + offset + 4, sizeof(ep));
+            uint32_t sp;
+            memcpy(&sp, content.data() + offset, sizeof(sp));
+            DEBUG_LOG("Adding entry_point_item: ep %08x, sp %08x\n", ep, sp);
+            entry_point = std::make_shared<entry_point_item>(ep, sp);
+            new_block.items.push_back(entry_point);
+        }
+    }
+
     hash_andor_sign(
         elf, &new_block, public_key, private_key,
         settings.seal.hash, settings.seal.sign,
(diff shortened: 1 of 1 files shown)


main.cpp
```diff
@@ -4594,6 +4594,29 @@ void sign_guts_elf(elf_file* elf, private_t private_key, public_t public_key) {
         new_block.items.push_back(version);
     }
 
+    // Add entry point when signing Arm images
+    std::shared_ptr<image_type_item> image_type = new_block.get_item<image_type_item>();
+    if (settings.seal.sign && image_type != nullptr && image_type->image_type() == type_exe && image_type->cpu() == cpu_arm) {
+        std::shared_ptr<entry_point_item> entry_point = new_block.get_item<entry_point_item>();
+        if (entry_point == nullptr) {
+            std::shared_ptr<vector_table_item> vtor = new_block.get_item<vector_table_item>();
+            uint32_t vtor_loc = 0x10000000;
```
kilograham, line 4603: should come from wherever the start of the image is; which in flash binaries is affected by roll; and would be in RAM for RAM binaries (where i guess it must be at 0x20000000 in these cases)

will-v-pi: Fixed - I've added support for adding the rolling window delta to the VTOR location, and it checks if the ELF entry point is in SRAM and reads the VTOR from there if that's the case

main.cpp
```diff
@@ -4594,6 +4594,34 @@ void sign_guts_elf(elf_file* elf, private_t private_key, public_t public_key) {
         new_block.items.push_back(version);
     }
 
+    // Add entry point when signing Arm images
+    std::shared_ptr<image_type_item> image_type = new_block.get_item<image_type_item>();
+    if (settings.seal.sign && image_type != nullptr && image_type->image_type() == type_exe && image_type->cpu() == cpu_arm) {
+        std::shared_ptr<entry_point_item> entry_point = new_block.get_item<entry_point_item>();
+        if (entry_point == nullptr) {
+            std::shared_ptr<vector_table_item> vtor = new_block.get_item<vector_table_item>();
+            uint32_t vtor_loc = elf->header().entry < SRAM_START ? 0x10000000 : 0x20000000;
```
kilograham, line 4603: do we need to check for XIP_SRAM-only binaries here? I forget if they can be signed (i believe so)

main.cpp
```diff
@@ -4630,6 +4658,27 @@ vector<uint8_t> sign_guts_bin(iostream_memory_access in, private_t private_key,
         new_block.items.push_back(version);
     }
 
+    // Add entry point when signing Arm images
```
kilograham, line 4661: Hmm; i hadn't really considered the BIN case where we don't know where the user is planning to load it. Perhaps we don't do all this and just give a warning instead? (or allow a "target address" option) - i'd be happy with the former for now, we can always add the new option later if it is useful

main.cpp
```diff
@@ -4594,6 +4594,34 @@ void sign_guts_elf(elf_file* elf, private_t private_key, public_t public_key) {
...
+            memcpy(&ep, content.data() + offset + 4, sizeof(ep));
+            uint32_t sp;
+            memcpy(&sp, content.data() + offset, sizeof(sp));
+            DEBUG_LOG("Adding entry_point_item: ep %08x, sp %08x\n", ep, sp);
+            entry_point = std::make_shared<entry_point_item>(ep, sp);
+            new_block.items.push_back(entry_point);
+        }
+    }
+
     hash_andor_sign(
```
lurch, line 4625: Screenshot from 2024-11-19 23-27-37

:wink:

main.cpp
will-v-pi: I've added support for this - although I'm not sure if we support XIP_SRAM binaries without a vector_table_item?

main.cpp
will-v-pi: I've changed it to use `bin_start`, which already comes from the `--offset` option that you can specify on the command line

main.cpp
kilograham: cool; note i looked at the code, but didn't check that the roll was in the right direction ;-)

kilograham (approved):

main.cpp
```diff
@@ -4607,6 +4607,41 @@ void sign_guts_elf(elf_file* elf, private_t private_key, public_t public_key) {
...
+                vtor_loc = 0x20000000;
+            } else if (elf->header().entry >= XIP_SRAM_START_RP2350) {
+                vtor_loc = 0x13ffc000;
+            } else {
+                vtor_loc = 0x10000000;
+            }
+            if (vtor != nullptr) {
+                vtor_loc = vtor->addr;
+            } else {
+                std::shared_ptr<rolling_window_delta_item> rwd = new_block.get_item<rolling_window_delta_item>();
```
kilograham, line 4627: note arguably rolling window delta only applies to flash binaries; silly to have it in another binary, but still we could move that into flash oly

kilograham (approved):

PR #167, opened 2024-11-07
Document that the user account needs to be in the plugdev group for udev rules to apply

MBradbury:
If a user follows the instructions in the README, they may not immediately identify that their user account needs to be in the plugdev group for the udev rules to apply. So be explicit and provide instructions to add the current user's account to the plugdev group.


will-v-pi:
I believe that this isn't the case due to the `uaccess` lines in the udev rules, which should allow the currently logged in user access even if they're not in the plugdev group - see #99 and #87

PR #168, opened 2024-11-07
lowercase 'Windows.h' in main.cpp for mingw32 support

dasimmet:
When trying to crosscompile with `ziglang.org` against ming32 libc I noticed the `Windows.h` header is not found.
On windows native it should not make a difference as the filesystem is case-insensitive.


will-v-pi:
Please target all PRs at the develop branch, not master

PR #170, opened 2024-11-08
Align Saving, Loading and Verifying progress bars

will-v-pi:
Align all the Saving, Loading and Verifying progress bars, by placing them at the longest possible prefix. Also prevent saving/loading from the RP2040 unstriped SRAM, as that returns an error from the device, and would increase the length of the longest possible prefix.

This changes the output from
```
$ picotool save -v flash.bin -r 0x10000000 0x10040000
Saving file: [==============================]  100%
Wrote 262144 bytes to flash.bin
Verifying Flash:    [==============================]  100%
  OK
$ picotool load -v flash.bin 
Loading into Flash: [==============================]  100%
Verifying Flash:    [==============================]  100%
  OK
```
to
```
$ picotool save -v flash.bin -r 0x10000000 0x10040000
Saving file:          [==============================]  100%
Wrote 262144 bytes to flash.bin
Verifying Flash:      [==============================]  100%
  OK
$ picotool load -v flash.bin 
Loading into Flash:   [==============================]  100%
Verifying Flash:      [==============================]  100%
  OK
```

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -3922,6 +3922,14 @@ static picoboot::connection get_single_rp2350_bootsel_device_connection(device_m
 
 struct progress_bar {
     explicit progress_bar(string prefix, int width = 30) : prefix(std::move(prefix)), width(width) {
+        // Align all bars with the longest possible prefix string
+        auto longest_mem = std::max_element(
+            std::begin(memory_names), std::end(memory_names),
+            [] (const auto & p1, const auto & p2) {
+                return p1.second.length() < p2.second.length();
+            }
+        );
+        longest_mem_str = longest_mem->second;
         progress(0);
     }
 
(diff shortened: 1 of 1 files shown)


main.cpp
```diff
@@ -3930,7 +3929,15 @@ struct progress_bar {
...
             percent = _percent;
             unsigned int len = (width * percent) / 100;
-            std::cout << prefix << "[" << string(len, '=') << string(width-len, ' ') << "]  " << std::to_string(percent) << "%\r" << std::flush;
+            // Align all bars with the longest possible prefix string
+            auto longest_mem = std::max_element(
+                std::begin(memory_names), std::end(memory_names),
+                [] (const auto & p1, const auto & p2) {
+                    return p1.second.length() < p2.second.length();
+                }
+            );
```
lurch, line 3938: Could / should this calculation be done just once, rather than for _every_ time you call the `progress()` function?

will-v-pi: Yep, I can calculate it once when the progress_bar is created

main.cpp
```diff
@@ -3922,21 +3922,22 @@ static picoboot::connection get_single_rp2350_bootsel_device_connection(device_m
...
             unsigned int len = (width * percent) / 100;
-            // Align all bars with the longest possible prefix string
-            auto longest_mem = std::max_element(
-                std::begin(memory_names), std::end(memory_names),
-                [] (const auto & p1, const auto & p2) {
-                    return p1.second.length() < p2.second.length();
-                }
-            );
-            string extra_space(string("Loading into " + longest_mem->second + ": ").length() - prefix.length(), ' ');
+            string extra_space(string("Loading into " + longest_mem_str + ": ").length() - prefix.length(), ' ');
```
lurch, line 3940: The `string("Loading into " + longest_mem_str + ": ").length()` could also be done just once? :stuck_out_tongue_winking_eye:

will-v-pi: Fixed by adding the extra space before setting the prefix

lurch: Excellent, cheers.

PR #173, opened 2024-11-12
Add all metadata blocks to info

will-v-pi:
This adds the `-m, --metadata` option to `picotool info` to print out details of all the metadata blocks in a binary.

Also standardise the indent for the whole `picotool info` command, so you don't get different sections with different indents.

For example, with the encrypted bootloader example it prints out:

```
$ picotool info -m enc_bootloader.uf2 
File enc_bootloader.uf2 family ID 'absolute':

Metadata Block 1
 address:               0x1000015c
 next block address:    0x10004e2c
 target chip:           RP2350
 image type:            ARM Secure
 vector table:          0x20070000

Metadata Block 2
 address:               0x10004e2c
 next block address:    0x10004ea0
 partition table:       non-singleton
 un-partitioned space:  S(rw) NSBOOT(rw) NS(rw), uf2 { absolute }
 partition 0 (A):       00010000->00080000 S(rw) NSBOOT(rw) NS(rw), id=0000000000000000, "A", uf2 { rp2350-arm-s }, arm_boot 1, riscv_boot 1
 partition 1 (B w/ 0):  00080000->000f0000 S(rw) NSBOOT(rw) NS(rw), id=0000000000000001, "B", uf2 { rp2350-arm-s }, arm_boot 1, riscv_boot 1
 version:               1.0

Metadata Block 3
 address:               0x10004ea0
 next block address:    0x1000015c
 target chip:           RP2350
 image type:            ARM Secure
 load map entry 0:      Clear 0x20000000+0x82000
 load map entry 1:      Copy 0x10000000+0x4e2c to 0x20070000
 load map entry 2:      Copy 0x10004e2c+0x74 to 0x20074e2c
 vector table:          0x20070000
 hash value:            verified
 signature:             verified
```

whereas the basic info only contains the metadata of the "best" block (the chosen block when performing an Arm boot)

```
$ picotool info enc_bootloader.uf2 
File enc_bootloader.uf2 family ID 'absolute':

Program Information
 name:              enc_bootloader
 web site:          https://github.com/raspberrypi/pico-examples/tree/HEAD/bootloaders/encrypted
 features:          UART stdin / stdout
 binary start:      0x10000000
 target chip:       RP2350
 image type:        ARM Secure
 load map entry 0:  Clear 0x20000000+0x82000
 load map entry 1:  Copy 0x10000000+0x4e2c to 0x20070000
 load map entry 2:  Copy 0x10004e2c+0x74 to 0x20074e2c
 vector table:      0x20070000
 hash value:        verified
 signature:         verified
```


will-v-pi:
> Are these "metadata blocks" equivalent to the "blocks" from the "block loop" described in section 5.1.5 of the RP2350 datasheet? I wonder if it'd also be useful to print the item_type of each block?

Yes, that’s what these blocks are - the item_type is per item in the block (eg vector_table_item, image_def_item), but I can add a print for the block type (image_def, partition_table or ignored)

lurch:
```
 load map entry 1:      Copy 0x10000000+0x4e2c to 0x20070000
```
So from what little I understand of section 5.9 of the RP2350 datasheet, I _think_ this means "copy 0x4e2c bytes from 0x10000000 to 0x20070000", but I guess a naive interpretation _might_ think that the `+` is indicating an offset, and so this line might be misinterpreted as "copy 0x10004e2c to 0x20070000" ?
The best non-ambiguous alternative I can come up with off the top of my head is:
```
 load map entry 1:      Copy 0x10000000:0x10004e2c to 0x20070000:0x20074e2c
```
to indicate that a range of bytes (which I assume is Flash) is being copied to some other range of bytes (which I assume is SRAM) ?

Similarly, perhaps
```
 load map entry 0:  Clear 0x20000000+0x82000
```
would be less ambiguous if written as
```
 load map entry 0:  Clear 0x20000000:0x20082000
```
? :man_shrugging:

lurch:
> > Are these "metadata blocks" equivalent to the "blocks" from the "block loop" described in section 5.1.5 of the RP2350 datasheet? I wonder if it'd also be useful to print the item_type of each block?
> 
> Yes, that’s what these blocks are - the item_type is per item in the block (eg vector_table_item, image_def_item), but I can add a print for the block type (image_def, partition_table or ignored)

Right, I guess I meant "block type" rather than "item type" :slightly_smiling_face:

will-v-pi:
> Ah, I've just noticed that you're already using `00010000->00080000` in the partition info, so for consistency perhaps it makes sense to use `->` instead of `:` in my "range" examples above. But also for consistency, perhaps it makes sense to add the `0x` prefixes to the addresses in the partition info!

I've improved the load_map printouts to be `0xstart->0xend` rather than `0xstart+0xsize` - I think I'll leave the partition printouts for now because those are flash addresses, whereas the load_map addresses are in the RP2xxx address space, so it's fine to have some differences between them

lurch:
> I've improved the load_map printouts to be `0xstart->0xend` rather than `0xstart+0xsize`

Just for reference, could you update the examples in the first post please? :slightly_smiling_face:

kilograham:
This is all very nice, but I think:

```
$ picotool info enc_bootloader.uf2 
File enc_bootloader.uf2 family ID 'absolute':

Program Information
 name:              enc_bootloader
 web site:          https://github.com/raspberrypi/pico-examples/tree/HEAD/bootloaders/encrypted
 features:          UART stdin / stdout
 binary start:      0x10000000
 target chip:       RP2350
 image type:        ARM Secure
 load map entry 0:  Clear 0x20000000->0x20082000
 load map entry 1:  Copy 0x10000000->0x10004e2c to 0x20070000->0x20074e2c
 load map entry 2:  Copy 0x10004e2c->0x10004ea0 to 0x20074e2c->0x20074ea0
 vector table:      0x20070000
 hash value:        verified
 signature:         verified
 ```
 
 load map/vector table should be guarded by `-a/--all` I think (possibly the hash/sig, though I'm happy with them if they are always succinct). Note, i just looked and i see they only have "verified/incorrect"; might be worth displaying something for present (or perhaps include the full hash/sig values on `-a`

**Comment by lurch** [comment]
I think I agree with Graham - probably makes sense to show the load map stuff when doing `picotool info -m` (as  you're explicitly asking for extra-info), but not when doing just a `picotool info` (which should just provide basic info).

**Comment by will-v-pi** [comment]
> Note, i just looked and i see they only have "verified/incorrect"; might be worth displaying something for present

It will only display that line if it's present, and then say verified or incorrect depending on whether the verification succeeded.

I can add verbose printout of the raw signature/hash bytes and public key when using `-m` or `-a`

**Comment by will-v-pi** [comment]
I think I've made all those changes - I've updated the description to the current state

File: main.cpp [hunk 1]
```diff
@@ -3103,60 +3103,85 @@ void info_guts(memory_access &raw_access, void *con) {
...
+                    info_pair("entry point", ss.str());
+                }
             }
 
             // Hash and Sig
             if (hash_verified != none) {
-                info_pair("hash value", hash_verified == passed ? "verified" : "incorrect");
+                info_pair("hash", hash_verified == passed ? "verified" : "incorrect");
+                if (verbose_metadata) {
+                    std::shared_ptr<hash_value_item> hash_value = best_block->get_item<hash_value_item>();
```
lurch, line 3160: Does the use of `best_block` mean that it'll always display the same hash when there are multiple IMAGE_DEFs in a block-loop, each with different hashes?

lurch:
> I've updated the description to the current state

Looks good! :raised_hands:

main.cpp
will-v-pi: No, `best_block` here is just the current block being printed - so for multiple IMAGE_DEFs it'll print separate hashes for each in their own metadata block printout (along with their separate load maps etc)

lurch: Ah cool - is it worth renaming the variable to `current_block` to avoid confusion?

lurch: Thank you! :heart_eyes:

kilograham:
cam you update the comment with `-a` too so we can see

will-v-pi:
> cam you update the comment with `-a` too so we can see

Added to the description - `-a` just prints the output of basic info combined with the `-m` output, plus the usual stuff

lurch:
Only a small nitpick, but I wonder if it might make sense to reorder these rows slightly:
```
 hash:                  verified
 hash value:            C3591E6E0AB0E0213803CADEF0F1CEA4CE2D887FCD47CDE2FB5271A957B994F8
 signature:             verified
 signature value:       561075BB5B6C962382CC5D7331C4ED5967D8E42BAD17643CCE46BA15FD3D49555E754DABDC7DF5A61318287F95B19C0929E44D23EB00B3F984013A0C9D0BC6E4
 public key:            61826D3101465B8001F78B54DEEFD08ABE6C45C62305832A09AFBC831B1877D3B0312DDDA962AC9D5073772CBB6C559D2735C331BF68DB4C5145DE7B43DC5263
```

so that you get the "verified" _after_ the values?
```
 hash value:            C3591E6E0AB0E0213803CADEF0F1CEA4CE2D887FCD47CDE2FB5271A957B994F8
 hash:                  verified
 signature value:       561075BB5B6C962382CC5D7331C4ED5967D8E42BAD17643CCE46BA15FD3D49555E754DABDC7DF5A61318287F95B19C0929E44D23EB00B3F984013A0C9D0BC6E4
 public key:            61826D3101465B8001F78B54DEEFD08ABE6C45C62305832A09AFBC831B1877D3B0312DDDA962AC9D5073772CBB6C559D2735C331BF68DB4C5145DE7B43DC5263
 signature:             verified
```

:shrug:

will-v-pi:
> Only a small nitpick, but I wonder if it might make sense to reorder these rows slightly:
> 
> ```
>  hash:                  verified
>  hash value:            C3591E6E0AB0E0213803CADEF0F1CEA4CE2D887FCD47CDE2FB5271A957B994F8
>  signature:             verified
>  signature value:       561075BB5B6C962382CC5D7331C4ED5967D8E42BAD17643CCE46BA15FD3D49555E754DABDC7DF5A61318287F95B19C0929E44D23EB00B3F984013A0C9D0BC6E4
>  public key:            61826D3101465B8001F78B54DEEFD08ABE6C45C62305832A09AFBC831B1877D3B0312DDDA962AC9D5073772CBB6C559D2735C331BF68DB4C5145DE7B43DC5263
> ```
> 
> so that you get the "verified" _after_ the values?
> 
> ```
>  hash value:            C3591E6E0AB0E0213803CADEF0F1CEA4CE2D887FCD47CDE2FB5271A957B994F8
>  hash:                  verified
>  signature value:       561075BB5B6C962382CC5D7331C4ED5967D8E42BAD17643CCE46BA15FD3D49555E754DABDC7DF5A61318287F95B19C0929E44D23EB00B3F984013A0C9D0BC6E4
>  public key:            61826D3101465B8001F78B54DEEFD08ABE6C45C62305832A09AFBC831B1877D3B0312DDDA962AC9D5073772CBB6C559D2735C331BF68DB4C5145DE7B43DC5263
>  signature:             verified
> ```
> 
> 🤷

I actually prefer it this way round - so you can see the important bit first (that the sig/hash is verified), and then the extra info comes after

kilograham:
i agree with above; one last nit though can we put the -m info after everything else; the info is kind of in the order of what user cares about most

will-v-pi:
> i agree with above; one last nit though can we put the -m info after everything else; the info is kind of in the order of what user cares about most

Done - I've moved it to just before device info

kilograham (approved):
nice

PR #175, opened 2024-11-12
Add option to ignore already-set bits in otp set command

### LINKED ISSUE #137: Add ability to OR values into non ECC rows
`picotool otp set` correctly disallows you from setting a value which would require clearing bits in the OTP; however often the intention is to just set additional bits; perhaps `picotool otp set_bits` or similar

will-v-pi:
This adds a way to set bits in an OTP row, without having to specify a field - for example for the page lock rows

Allows you to workaround this (for example)
```bash
$ picotool otp set crit1 0x1
ROW 0x0040  OLD_VALUE=0x000000: OTP_DATA_CRIT1
        "Page 1 critical boot flags (RBIT-8)"
$ picotool otp set crit1 0x2
ROW 0x0040  OLD_VALUE=0x000001: OTP_DATA_CRIT1
        "Page 1 critical boot flags (RBIT-8)"
ERROR: Cannot clear bits in OTP row(s): current value 000001, new value 000002
```
using the `-s` argument
```bash
$ picotool otp set crit1 0x2 -s
ROW 0x0040  OLD_VALUE=0x000001: OTP_DATA_CRIT1
        "Page 1 critical boot flags (RBIT-8)"
$ picotool otp get crit1
ROW 0x0040: OTP_DATA_CRIT1
        "Page 1 critical boot flags (RBIT-8)"

    VALUE 0x000003

    field GLITCH_DETECTOR_SENS (bits 5-6) = 0
        "Increase the sensitivity of the glitch detectors from their default."
    field GLITCH_DETECTOR_ENABLE (bit 4) = 0
        "Arm the glitch detectors to reset the system if an abnormal clock/power event is observed."
    field BOOT_ARCH (bit 3) = 0
        "Set the default boot architecture, 0=ARM 1=RISC-V. Ignored if ARM_DISABLE, RISCV_DISABLE or SECURE_BOOT_ENABLE is set."
    field DEBUG_DISABLE (bit 2) = 0
        "Disable all debug access"
    field SECURE_DEBUG_DISABLE (bit 1) = 1
        "Disable Secure debug access"
    field SECURE_BOOT_ENABLE (bit 0) = 1
        "Enable boot signature enforcement, and permanently disable the RISC-V cores."
```

Fixes #137


lurch:
```
ERROR: Cannot clear bits in OTP row(s): current value 000001, new value 000002
```
I wonder if this error message should point people at the `-s` option?

Also, this PR should probably update https://github.com/raspberrypi/picotool/tree/develop?tab=readme-ov-file#otp too?

will-v-pi:
> 's' is OK, but I think `ignore-set` is confusing... I think `set-bits-only` (or probably just `set-bits`)-is better, and we should probably disallow it with ECC

I've changed it to `set-bits` - although I've left the internal variable as `ignore_set` as that's what picotool's doing internally

It already fails on ECC rows in the normal way: https://github.com/raspberrypi/picotool/blob/fb85aca4cf49e2ad998db4fd9f9160b00893b663/main.cpp#L7107-L7109

kilograham (approved):
Approving, but 

> I've changed it to set-bits - although I've left the internal variable as ignore_set as that's what picotool's doing internally

Seems like a weird hill to die on... it doesn't ignore anything; it ORs the old value into the new value ;-)

PR #176, opened 2024-11-15
Add JSON schemas

### LINKED ISSUE #172: Document schema for all JSON file arguments
Please consider documenting what the schema is for all commands that take a JSON file.

While there are example JSON for some of them (partitions, white label, permissions), it would be ideal all JSON schema for each such argument was documented with a description of what each field in the JSON means, and what its possible values are.

will-v-pi:
Add JSON schemas for the JSON inputs used.

Currently the URLs point to the jsonschema branch - these should be changed to the develop branch once merged.

Fixes #172


README.md
```diff
@@ -952,7 +952,7 @@ $ picotool reboot
 ### white-label
 
 This command allows for OTP white-labelling, which sets the USB configuration used by the device in BOOTSEL mode.
-This can be configured from a JSON file, an example of which is in [sample-wl.json](sample-wl.json).
+This can be configured from a JSON file, an example of which is in [sample-wl.json](sample-wl.json). The schema for this JSON file is [here](json/schemas/whitelabel-schema.json)
```
lurch, line 955: It looks like this PR also moves `sample-wl.json` to `json/sample-wl.json` ? (and same comment for the other "sample" JSON files too)

json/default-pt.json
```diff
@@ -0,0 +1,37 @@
+{
+    "$schema": "https://raw.githubusercontent.com/raspberrypi/picotool/jsonschema/json/schemas/partition-table-schema.json",
```
lurch, line 2: Should we ensure that `rp2350` is in the schema-URL somewhere, in case future chips need different schema files?

json/schemas/otp-contents-schema.json
```diff
@@ -0,0 +1,77 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "OTP Contents",
+    "description": "Defined contents of the OTP",
```
lurch, line 4: "...of the RP2350 OTP" ?

json/schemas/partition-table-schema.json
```diff
@@ -0,0 +1,175 @@
...
+                        },
+                        "bootloader": {
+                            "description": "Bootloader Permissions",
+                            "type": "string",
+                            "pattern": "^(r|w){0,2}$"
+                        }
+                    }
+                }
+            },
+            "required": ["permissions", "families"],
```
lurch, line 64: If `permissions` is required, are the sub-properties within `permissions` also required?

will-v-pi: No, just an empty permissions will work fine
```
"permissions": {}
```

json/schemas/partition-table-schema.json
```diff
@@ -0,0 +1,175 @@
...
+                                "data",
+                                "absolute",
+                                "rp2040",
+                                "rp2350-arm-s",
+                                "rp2350-arm-ns",
+                                "rp2350-riscv"
+                            ]
+                        }
+                    },
+                    "permissions": {
```
lurch, line 117: It looks like this `permissions` part of the schema is identical to the earlier `permissions` part? Might be worth using https://json-schema.org/understanding-json-schema/structuring#defs to avoid repetition.

json/schemas/whitelabel-schema.json
```diff
@@ -0,0 +1,119 @@
...
+                    "type": "string",
+                    "maxLength": 30
+                },
+                "serial_number": {
+                    "description": "Serial Number (can contain unicode)",
+                    "type": "string",
+                    "maxLength": 30
+                },
+                "max_power": {
+                    "description": "Max power consumption. in 2mA units",
```
lurch, line 49: nitpick: I guess either the `.` should be a `,`, or the `In` should be capitalised?

lurch:
Should one of these schema files contain all the info from section 13.9 of the RP2350 datasheet, or is this the wrong place for that?

will-v-pi:
> Should one of these schema files contain all the info from section 13.9 of the RP2350 datasheet, or is this the wrong place for that?

I think this is the wrong place for that - that section of the datasheet is probably the best place for that information, as `picotool` should just handle all the redundancy settings etc for the predefined values without issue when reading/writing OTP.

json/schemas/otp-contents-schema.json
will-v-pi: I'll go with "... of the RP-series device OTP"

json/default-pt.json
will-v-pi: I think the schema would hopefully stay the same for future chips - this isn't really a proper URL anyway, if we wanted to do it properly then we could add GitHub Pages to host the schemas, and then host schemas for each released version of picotool (in case they do change at a later date)

json/schemas/whitelabel-schema.json
will-v-pi: Typo - fixed

json/default-pt.json
will-v-pi: Actually, we probably don't need full on GitHub Pages for this - the `raw.githubusercontent` URLs have versioning, so you can just use URLs like this:
```
https://raw.githubusercontent.com/raspberrypi/picotool/master/json/schemas/partition-table-schema.json
https://raw.githubusercontent.com/raspberrypi/picotool/develop/json/schemas/partition-table-schema.json
https://raw.githubusercontent.com/raspberrypi/picotool/2.1.0/json/schemas/partition-table-schema.json
```

I think this is probably the simplest solution - you can stick to a specific version, or track `master` or `develop`

lurch: > I think the schema would hopefully stay the same for future chips

Ahhh, fair enough. Is that likely to be true for **all** the schemas that you're adding in this PR?

json/schemas/whitelabel-schema.json
```diff
@@ -0,0 +1,119 @@
...
+                    "maxLength": 30
+                },
+                "max_power": {
+                    "description": "Max power consumption, in 2mA units",
+                    "type": ["integer", "string"],
+                    "maximum": 255,
+                    "pattern": "^0x[0-9a-fA-F]{0,2}$"
+                },
+                "attributes": {
+                    "description": "Device attributes: bit 7 must be 1, bit 6 is self-powered, bit 5 is remote wakeup, bits 0-4 must be 0",
```
lurch, line 55: If bit 7 must be 1, does that imply that 128 is the minimum value? And if bits 0-4 must be 0, then I guess 255 isn't the maximum value?

json/schemas/whitelabel-schema.json
```diff
@@ -0,0 +1,119 @@
...
+                    "type": ["integer", "string"],
+                    "maximum": 255,
+                    "pattern": "^0x[0-9a-fA-F]{0,2}$"
```
lurch, line 58: If bit 7 must be 1, then I guess the pattern here should use `{2}` rather than `{0,2}` ?

json/schemas/whitelabel-schema.json
```diff
@@ -0,0 +1,119 @@
...
+                "serial_number": {
+                    "description": "Serial Number (can contain unicode)",
+                    "type": "string",
+                    "maxLength": 30
+                },
+                "max_power": {
+                    "description": "Max power consumption, in 2mA units",
+                    "type": ["integer", "string"],
+                    "maximum": 255,
+                    "pattern": "^0x[0-9a-fA-F]{0,2}$"
```
lurch, line 52: Should the regex here use `{1,2}` rather than `{0,2}`? Because it seems like `0x` wouldn't be a valid value?

json/schemas/whitelabel-schema.json
will-v-pi: Fixed - the range is 128-224. Technically only the 4 values `0x80`, `0xA0`, `0xC0` and `0xE0` are allowed, so I've changed the regex to match that

json/schemas/whitelabel-schema.json
will-v-pi: Yep - fixed

json/schemas/otp-schema.json
```diff
@@ -0,0 +1,50 @@
...
+            "properties": {
+                "ecc": {
+                    "description": "Protect with ECC",
+                    "type": "boolean"
+                },
+                "value": {
+                    "description": "Value to write",
+                    "type": ["array", "string", "integer"],
+                    "pattern": "^0x[0-9a-fA-F]{1,6}$",
+                    "items": {
```
lurch, line 20: I guess `items` here could also be a `$ref`, but I'll let you decide if it's worth it or not.

json/default-pt.json
will-v-pi: > > I think the schema would hopefully stay the same for future chips
> 
> Ahhh, fair enough. Is that likely to be true for **all** the schemas that you're adding in this PR?

The hope would be that they will all be backwards compatible at least - but as I said if they do change then you can pin to a specific version using the URL, without needing any special folder structure.

kilograham (approved):

lurch:
@will-v-pi Don't forget to update the schema URLs before hitting the big green Merge button :smile:

PR #181, opened 2024-11-22
Preemptively update Bazel dep on Pico SDK for 2.1.0 release

armandomontanez:
While pico-sdk@2.1.0 will be backwards compatible with picotool@2.0.0, picotool@2.1.0 will require pico-sdk@2.1.0 to properly build.


kilograham:
was fixed out of band

Issue #187, opened 2024-12-08
metadata.h:258:44: warning: multiple unsequenced modifications to 'i' [-Wunsequenced]


ericm1024:
When building a pico example project, I receive the following warning compiling picotool
```
In file included from /Users/eric/Desktop/pdu/git/pico/build/_deps/picotool-src/bintool/bintool.cpp:22:                                                                                                         
In file included from /Users/eric/Desktop/pdu/git/pico/build/_deps/picotool-src/bintool/bintool.h:9:                                                                                                            
/Users/eric/Desktop/pdu/git/pico/build/_deps/picotool-src/bintool/metadata.h:258:44: warning: multiple unsequenced modifications to 'i' [-Wunsequenced]
                new_p.id = (uint64_t)data[i++] | ((uint64_t)data[i++] << 32);                                                                                                                                   
                                           ^                      ~~                                                                                                                                            
/Users/eric/Desktop/pdu/git/pico/build/_deps/picotool-src/bintool/metadata.h:258:44: warning: multiple unsequenced modifications to 'i' [-Wunsequenced]
                new_p.id = (uint64_t)data[i++] | ((uint64_t)data[i++] << 32);                                                                                                                                   
                                           ^                      ~~                                                                                                                                            
/Users/eric/Desktop/pdu/git/pico/build/_deps/picotool-src/bintool/metadata.h:624:47: note: in instantiation of function template specialization 'partition_table_item::parse<std::__wrap_iter<unsigned int *>>' 
requested here                                                                                                                                                                                                  
                    i = partition_table_item::parse(it, end, header);                                                                                                                                           
                                              ^                                                                                                                                                                 
/Users/eric/Desktop/pdu/git/pico/build/_deps/picotool-src/bintool/bintool.cpp:121:50: note: in instantiation of function template specialization 'block::parse<std::__wrap_iter<unsigned int *>>' requested here
                            first_block = block::parse(x->physical_address() + 4*(block_begin - words.begin() - 1),                      
                                                 ^                                                                                                                                                              
2 warnings generated.                                                                                                                                                                                           
[ 92%] Built target bintool
```
Compiler is
```
eric@Erics-Air build % /Library/Developer/CommandLineTools/usr/bin/c++ --version
Apple clang version 15.0.0 (clang-1500.3.9.4)
Target: arm64-apple-darwin23.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

[...]

PR #188, opened 2024-12-08
Don't rely on unsequenced evaluation order in bintool/metadata.h (#187)

ericm1024:
With the following compiler:
% /Library/Developer/CommandLineTools/usr/bin/c++ --version Apple clang version 15.0.0 (clang-1500.3.9.4)
Target: arm64-apple-darwin23.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

I was getting the following warning:

metadata.h:258:44: warning: multiple unsequenced modifications to 'i' [-Wunsequenced]

It appears as though the code relies on the leftmost i++ happening first, followed by the rightmost, but the language seemingly provides no such guarantee.

Fix this by reading the low and high words in order, then combining them, rather than doing so in a single expression.

### DIFF
--- a/bintool/metadata.h
+++ b/bintool/metadata.h
@@ -255,7 +255,9 @@ struct partition_table_item : public single_byte_size_item {
             new_p.flags = permissions_flags & (~PICOBIN_PARTITION_PERMISSIONS_BITS);
 
             if (new_p.flags & PICOBIN_PARTITION_FLAGS_HAS_ID_BITS) {
-                new_p.id = (uint64_t)data[i++] | ((uint64_t)data[i++] << 32);
+                uint32_t low = data[i++];
+                uint32_t high = data[i++];
+                new_p.id = (uint64_t)low | ((uint64_t)high << 32);
             }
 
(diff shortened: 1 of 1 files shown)


bintool/metadata.h
```diff
@@ -255,7 +255,9 @@ struct partition_table_item : public single_byte_size_item {
             new_p.flags = permissions_flags & (~PICOBIN_PARTITION_PERMISSIONS_BITS);
 
             if (new_p.flags & PICOBIN_PARTITION_FLAGS_HAS_ID_BITS) {
-                new_p.id = (uint64_t)data[i++] | ((uint64_t)data[i++] << 32);
+                uint32_t low = data[i++];
+                uint32_t high = data[i++];
+                new_p.id = (uint64_t)low | ((uint64_t)high << 32);
```
lurch, line 260: Alternatively, I guess you could do
```c
new_p.id = (uint64_t)data[i] | ((uint64_t)data[i+1] << 32);
i += 2;
```
which would avoid the temporary variables?

lurch:
Oh, and can you change the base branch to `develop` please, instead of `master`?

bintool/metadata.h
will-v-pi: I think I prefer the original style - using multiple `i++` matches the style used throughout the rest of the file

PR #191, opened 2024-12-18
Add Block Device Support

### LINKED ISSUE #58: Add support for read/writing files on embedded drives
picotool is able to find embedded drives; it might be nice to be able to list/extract/and writes files

will-v-pi:
Adds support for using `ls`, `mkdir`, `cp` (including between host and device), `rm`, and `cat` commands, with LittleFS (used by MicroPython) or FatFS filesystem support.

Works with block devices specified in the binary info, and with partitions. It autodetects the file system if `--filesystem` is not specified, and also supports formatting the block device if required with the `-f` argument.

This fixes #58


README.md
```diff
@@ -1392,6 +1394,265 @@ OPTIONS:
...
+$ picotool help bdev ls
+BDEV LS:
+    List contents of the block device
+
+SYNOPSIS:
+    picotool bdev ls <dirname> [-r] [-p <partition>] [--filesystem <fs>] [-f] [device-selection]
+
+OPTIONS:
+        <dirname>
+            The file name
```
lurch: "dirname" vs. "file name" ? :wink: 
Also, if I wanted to list the "root directory" would I give it a `<dirname>` of `/`, or something else?

README.md
```diff
@@ -1392,6 +1394,265 @@ OPTIONS:
...
+$ picotool help bdev mkdir
+BDEV MKDIR:
+    Create directory on the block device
+
+SYNOPSIS:
+    picotool bdev mkdir <dirname> [-p <partition>] [--filesystem <fs>] [-f] [device-selection]
+
+OPTIONS:
+        <dirname>
+            The file name
```
```suggestion
        <dirname>
            The directory name
```

README.md
```diff
@@ -1392,6 +1394,265 @@ OPTIONS:
...
+    Block device options
+        -p, --partition
+            Partition to use as block device
+        <partition>
+            partition number
+        --filesystem
+            Specify filesystem to use
+        <fs>
+            littlefs|fatfs
+        -f, --format
```
lurch: Hmmm, we now seem to have **BOTH** `-f, --format` and `-f, --force` flags?! Either this worn't work properly, or if it does work it's pretty confusing!

README.md
```diff
@@ -1392,6 +1394,265 @@ OPTIONS:
...
+            command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
+        -F, --force-no-reboot
+            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
+            command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but without the
+            USB drive mounted
+```
+
+### cp
+
+Copy file to/from the block device - use :filename to indicate files on the device (eg `cp main.py :main.py` to upload to the device)
```
lurch: Probably doesn't hurt to be explicit?
```suggestion
Copy file to/from the block device - use :filename to indicate files on the device (e.g. `cp main.py :main.py` to upload to the device, or `cp :main.py main.py` to download from the device)
```

lurch: * Does this `bdev cp` command check that one (and only one) of either `src` or `dest` starts with a colon?
* Does it only work for files and not directories?
* If the answer to the previous question is "only files", would it be a lot of additional work to be able to copy directories? (which would presumably need to operate recursively)

lib/littlefs/README.md
```diff
@@ -0,0 +1,7 @@
+littlefs library
+================
+
+The upstream source for the files in this directory is
+https://github.com/littlefs-project/littlefs
+
+The current files come from v2.8.2
```
lurch: _Maybe_ worth bumping (after testing, of course) to the latest release? https://github.com/littlefs-project/littlefs/releases/tag/v2.11.1

main.cpp
```diff
@@ -786,6 +829,115 @@ struct config_command : public cmd {
...
+    (option('f', "--format").set(settings.bdev.format) % "Format the drive if necessary (may result in data loss)")
+).min(0).doc_non_optional(true) % "Block device options";
+
+struct bdev_ls_command : public cmd {
+    bdev_ls_command() : cmd("ls") {}
+    bool execute(device_map& devices) override;
+
+    group get_cli() override {
+        return (
+            optional_untyped_file_selection_x("dirname", 0) +
```
lurch: Does the `optional_` here indicate that this is an optional argument? If so, it should probably appear in the `README.md` as `[dirname]` rather than `<dirname>` ?

main.cpp
```diff
@@ -2205,34 +2360,45 @@ struct picoboot_memory_access : public memory_access {
...
+                // Check what's there already
+                write_data.resize(size);
+                read(address, write_data.data(), size, false);
+                // Check if we even need to write
+                if (std::equal(write_data.cbegin(), write_data.cend(), buffer)) {
+                    return;
+                }
+                // Check if we need to erase (ie check for non 0xff)
+                if (!std::all_of(write_data.cbegin(), write_data.cend(), [](uint8_t v) { return v == 0xff; })) {
+                    // Do automatically erase flash, and make it aligned
```
```suggestion
                    // To automatically erase flash, and make it aligned
```
?

main.cpp
```diff
@@ -2205,34 +2360,45 @@ struct picoboot_memory_access : public memory_access {
...
+            // Flash Translation Layer - auto-erase, and only write changed data
+            if (enable_ftl) {
+                // Check what's there already
+                write_data.resize(size);
+                read(address, write_data.data(), size, false);
+                // Check if we even need to write
+                if (std::equal(write_data.cbegin(), write_data.cend(), buffer)) {
+                    return;
+                }
+                // Check if we need to erase (ie check for non 0xff)
```
lurch: AIUI, when you erase Flash it sets it to all-1s, and so when you "write data" to Flash, it only flips some of the 1s to 0s. So I guess that _theoretically_, if you're overwriting `0b11111110` with `0b11111100` then there's also no need to erase the Flash first? Which suggests we could do something more sophisticated here than just "check for non 0xff" ? :thinking:

README.md
will-v-pi: Just pass no `dirname`, like you would to the normal `ls` command - it’s an optional argument

README.md
will-v-pi: Good spot - I’ll make this filesystem one not have a short argument

README.md
will-v-pi: - No - you can copy a file to somewhere else on the device, or even use this to copy local files if you really wanted
- Yes, it only works for files
- Could probably add that - just needs a recursive walk like `ls` has

main.cpp
will-v-pi: Yes, it’s optional - I’m not sure why the CLI stuff isn’t putting square brackets around it

main.cpp
will-v-pi: The comments were meant to mean “Check if we want to erase” followed by “Yes, we do want erase” - so maybe just delete the “Do”?
```suggestion
                    // Automatically erase flash, and make it aligned
```

main.cpp
will-v-pi: That’s true, I could probably check for `write_data & ~current_data` instead

main.cpp
```diff
@@ -5913,6 +6090,738 @@ bool link_command::execute(device_map &devices) {
...
+    if ((char)(settings.filenames[0].front()) == ':') {
+        srcRemote = true;
+        settings.filenames[0].erase(0, 1);
+    }
+    if ((char)(settings.filenames[1].front()) == ':') {
+        destRemote = true;
+        settings.filenames[1].erase(0, 1);
+    }
+
+    fos << "Copying " << (srcRemote ? "device " : "local ") << settings.filenames[0] << " to " << (destRemote ? "device " : "local ") << settings.filenames[1] << "\n";
```
lurch, line 6975: Perhaps it's worth checking that `(srcRemote || destRemote)`, as there seems to be little point using `picotool bdev cp` to copy one local file to another local file? :wink:

main.cpp
```diff
@@ -5913,6 +6090,738 @@ bool link_command::execute(device_map &devices) {
...
+
+    std::vector<char> data_buf;
+
+    // Read local file
+    if (!srcRemote) {
+        auto infile = get_file_idx(ios::in|ios::binary|ios::ate, 0);
+        auto size = infile->tellg();
+        infile->seekg(0, std::ios::beg);
+        data_buf.resize(size);
```
lurch, line 6989: If you accidentally point this at the wrong local filename (e.g. a huge disk-image file), is there a risk that this might cause an out-of-memory error?

README.md
```diff
@@ -1392,6 +1394,265 @@ OPTIONS:
...
+            command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
+        -F, --force-no-reboot
+            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
+            command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but without the
+            USB drive mounted
+```
+
+### cat
+
+Print contents of file on the block device
```
lurch, line 1975: Will an appropriate error get displayed if you try to `bdev cat` a directory name?

README.md
lurch: (perhaps the arg-parsing library could/should raise an error if given the same argument multiple times?)

lurch:
A thought I just had - to make this slightly easier to use, could you have `picotool` check what filename it was invoked with (e.g. like `busybox` does), and then you could have `picotool_ls`, `picotool_cat`, etc. symlinks to the `picotool` binary, and then calling `picotool_ls somedir` would do the same thing as `picotool bdev ls somedir` ? (Or if that's not possible, perhaps the picotool CMake install could add a bunch of shell scripts e.g. named `picotool_ls` which just does `picotool bdev ls $@`)

main.cpp
lurch: Ahhh, I didn't realise that
```c
// Do automatically erase flash, and make it aligned
// we have to erase in whole pages
```
was intended to be two separate sentences, and not one long sentence :slightly_smiling_face:

main.cpp
will-v-pi: I don’t think there’s any point removing that functionality, in case there’s a use case where you just use `picotool bdev cp` for everything - and this message makes it clear what’s happening when you do that, so if you didn’t mean to you can add a colon

main.cpp
will-v-pi: Yes - should probably check the input file fits in the largest flash size (32MB currently)

lurch: That raises an interesting point - are you able to calculate how much "free space" there is in the block device, and what happens if you try writing a file larger than this? Hopefully it doesn't just continue writing to Flash after the end of the block device?

will-v-pi:
> A thought I just had - to make this slightly easier to use, could you have `picotool` check what filename it was invoked with (e.g. like `busybox` does), and then you could have `picotool_ls`, `picotool_cat`, etc. symlinks to the `picotool` binary, and then calling `picotool_ls somedir` would do the same thing as `picotool bdev ls somedir` ? (Or if that's not possible, perhaps the picotool CMake install could add a bunch of shell scripts e.g. named `picotool_ls` which just does `picotool bdev ls $@`)

I think that might be best left to the user to do, to avoid polluting their shell (eg you might want `picotool_ls` to list devices, not use the `bdev` command) - you could also create an alias to `picotool bdev` in bashrc (which also avoids the need for underscores and therefore the shift key), so I don’t think I’ll include that is this PR

main.cpp
will-v-pi: That's up to the filesystem - both LittleFS and FatFS should throw an error if there isn't enough free space on the device, but I might add a backstop to prevent writing past the end of the block device too

README.md
will-v-pi: > (perhaps the arg-parsing library could/should raise an error if given the same argument multiple times?)

I will leave that for a future PR

README.md
will-v-pi: I've pushed a change to print error strings instead of numbers, so you'd get `ERROR: LittleFS Open Error: Entry is a dir`

main.cpp
```diff
@@ -6613,6 +6647,11 @@ bool bdev_cp_command::execute(device_map &devices) {
     if (!srcRemote) {
         auto infile = get_file_idx(ios::in|ios::binary|ios::ate, 0);
         auto size = infile->tellg();
+        uint32_t max_size = bdevfs_setup.access->get_model()->flash_end() - bdevfs_setup.access->get_model()->flash_start();
+        if (size > max_size) {
+            fail(ERROR_NOT_POSSIBLE, "File size %dMB is too large for flash device (max %dMB)", size / (1024 * 1024), max_size / (1024 * 1024));
```
lurch, line 6985: Is it possible for this to (also?) check against the "block device" size, rather than just the "flash device" size? Or would that be a premature optimisation because it'll get caught by later checks anyway?

README.md
```diff
@@ -1392,6 +1394,266 @@ OPTIONS:
...
+        <dirname>
+            The directory name to list (optional)
+        -r, --recursive
+            List files in directories recursively
+    Block device options
+        -p, --partition
+            Partition to use as block device
+        <partition>
+            partition number
+        --filesystem
```
lurch, line 1731: Is the `--filesystem` flag only needed when formatting a block-device, or are there likely to be circumstances in which the filesystem auto-deteciton doesn't work?

README.md
```diff
@@ -1392,6 +1394,266 @@ OPTIONS:
...
+            Specify filesystem to use
+        <fs>
+            littlefs|fatfs
+        --format
+            Format the drive if necessary (may result in data loss)
```
lurch, line 1740: Presumably using the `--format` option in conjunction with `picotool bdev ls` will **always** return an empty directory listing? :joy:  But this makes me wonder why `--format` is an option that can be used in combination  with any other `picotool bdev` command, rather than just being a standalone `picotool bdev format` or `picotool format_bdev` command? :thinking:

README.md
```diff
@@ -1392,6 +1394,266 @@ OPTIONS:
...
+```
+
+### cat
+
+Print contents of file on the block device
+
+```text
+$ picotool help bdev cat
+BDEV CAT:
+    Print contents of file on the block device
```
lurch, line 1980: So I guess `picotool bdev cat main.py > main.py` effectively does the same thing as `picotool bdev cp :main.py main.py` ?

main.cpp
```diff
@@ -94,6 +97,10 @@ static __forceinline int __builtin_ctz(unsigned x) {
 #ifndef BOOTROM_FAMILY_ID_MAX
 #define BOOTROM_FAMILY_ID_MAX       RP2350_ARM_NS_FAMILY_ID
 #endif
+#ifndef BLOCK_DEVICE_PARTITION_ID
+// The default 0x626C6F636B646576 value is the ASCII encoding of "blockdev"
+#define BLOCK_DEVICE_PARTITION_ID 0x626C6F636B646576
```
lurch: Should this ID also appear somewhere in a pico-sdk header, in case 3rd-party utilities also want to use it? Or do we expect any 3rd-party utilities to just copy'n'paste this into their own source code?

main.cpp
```diff
@@ -2205,34 +2364,52 @@ struct picoboot_memory_access : public memory_access {
...
+            // Flash Translation Layer - auto-erase, and only write changed data
+            if (enable_ftl) {
+                // Check what's there already
+                write_data.resize(size);
+                read(address, write_data.data(), size, false);
+                // Check if we even need to write
+                if (std::equal(write_data.cbegin(), write_data.cend(), buffer)) {
+                    return;
+                }
+                // Check if we need to erase (ie check for non 0xff)
```
lurch: Does this comment need to be updated?

main.cpp
```diff
@@ -5913,6 +6101,820 @@ bool link_command::execute(device_map &devices) {
...
+        bdevfs_setup.writeable = std::get<2>((*partitions)[settings.bdev.partition]) & PICOBIN_PARTITION_PERMISSION_NSBOOT_W_BITS;
+        start += FLASH_START;
+        end += FLASH_START;
+        if (end <= start) {
+            fail(ERROR_ARGS, "Partition range is invalid/empty");
+        }
+        bdevfs_setup.base_addr = start;
+        bdevfs_setup.size = end - start;
+        block_device_found = true;
+    } else if (partitions) { // use first partition with id "blockdev"
```
lurch: It's probably a bit of an extreme edge-case, but does there need to be a mechanism for ignoring a partition named "blockdev", if you happen to have a partition with that name, but you still want to use the block device specified in the binary info?

main.cpp
```diff
@@ -5913,6 +6101,820 @@ bool link_command::execute(device_map &devices) {
...
+        res = strcmp(littlefs_str, "littlefs");
+        if (res == 0) {
+            // LittleFS Found
+            fos << "LittleFS Filesystem Detected\n";
+            settings.bdev.fs = fs_littlefs;
+            return;
+        }
+
+        // No FS Found
+        fail(ERROR_CONNECTION, "No file system detected - to format the drive use `-f --filesystem <littlefs|fatfs>`");
```
lurch: I guess the `-f` here needs to be changed to `--format`?

lib/littlefs/README.md
will-v-pi: Bumped, and seems to work fine

main.cpp
will-v-pi: That'll get caught by later checks by the filesystem anyway - this check is just to prevent reading a multi-gigabyte file into memory

README.md
will-v-pi: With the current supported filesystems it should only be needed when formatting, but future filesystems may be harder to auto-detect so may require this flag

README.md
will-v-pi: I added it as a flag because of where the formatting occurs in the code - it will only format the drive if no filesystem (of the type specified by `--filesystem` if that is passed) is detected, so you could always call your first command with `--format` if you wanted to make sure you always do the command. Eg this works for both formatting new devices and updating the file on old devices:
```
picotool bdev cp main.py :/ --format --filesystem littlefs
```

README.md
will-v-pi: Yes - this also allows for things like `picotool bdev cat main.py | sha256sum`

main.cpp
will-v-pi: Yes, if it's decided that this is wanted as a way to identify block device partitions, then I'll open an SDK PR to add this to the SDK

main.cpp
will-v-pi: It's not a partition named `blockdev` (the partition name can be anything), it's a partition with the 64-bit ID of the ascii characters `blockdev`, which I think is an intentional enough thing to do that we don't need to add handling for ignoring it?

README.md
lurch: Oh! I see now that I'd misread "Format the drive if necessary" as just "Format the drive" :facepalm: 

So I guess that means if you had a blockdevice that was already fatfs-formatted, and you wanted to explicitly reformat it, you'd need to do something like:
```
picotool bdev ls --format --filesystem littlefs
picotool bdev ls --format --filesystem fatfs
```
?

will-v-pi: Yes, so it probably is worth adding a separate `picotool bdev format` command to unconditionally format the block device

main.cpp
kilograham: sounds good

main.cpp
```diff
@@ -6081,6 +6269,820 @@ bool link_command::execute(device_map &devices) {
...
+        end += FLASH_START;
+        if (end <= start) {
+            fail(ERROR_ARGS, "Partition range is invalid/empty");
+        }
+        bdevfs_setup.base_addr = start;
+        bdevfs_setup.size = end - start;
+        block_device_found = true;
+    } else if (partitions) { // use first partition with id "blockdev"
+        for (auto &partition : *partitions) {
+            if (std::get<3>(partition) != BLOCK_DEVICE_PARTITION_ID) {
```
kilograham: seems a bit mean to not let them specify the partition id (or name)

main.cpp
```diff
@@ -6081,6 +6269,820 @@ bool link_command::execute(device_map &devices) {
...
+        if (has_binary_info) {
+            auto access = remapped_memory_access(*bi_access, hdr.reverse_copy_mapping);
+            auto visitor = bi_visitor{};
+            bool block_device_found = false;
+            visitor.block_device([&](memory_access &access, binary_info_block_device_t &bi_bdev) {
+                block_device_found = true;
+                std::stringstream ss;
+                ss << hex_string(bi_bdev.address) << "-" << hex_string(bi_bdev.address + bi_bdev.size) <<
+                    " (" << ((bi_bdev.size + 1023) / 1024) << "K): " << read_string(access, bi_bdev.name);
+                if (bi_bdev.flags) {
```
kilograham, line 6426: ditto can we let them select

README.md
```diff
@@ -1660,6 +1662,306 @@ OPTIONS:
             The file name
 ```
 
+## bdev
+
+The `bdev` commands are for interacting with block devices in Flash. These can either be defined using binary info, or you can specify a partition to use as a block device. They work with a LittleFS filesystem, or a FatFS filesystem.
```
kilograham: These can be defined is confusing; who is defining what, when? perhaps you mean selected not defined. Also perhaps "the commands" rather than They

README.md
```diff
@@ -1660,6 +1662,306 @@ OPTIONS:
...
+            Assume the device is an RP2040 - this is only required when using a custom vid/pid with an RP2040 on Windows, and is ignored on
+            other operating systems
+        -f, --force
+            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
+            command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
+        -F, --force-no-reboot
+            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
+            command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but without the
+            USB drive mounted
+        --bootsel-led <gpio>
```
kilograham, line 1762: probably a minor grumble, and maybe these are mostly applicable to all commands as you can force a reboot, and therefore there can be an activity LED, so maybe its fine... its just  abit weird that it is in target defice selection

kilograham (changes requested):
might be worth adding some tests to make sure none of the commands delete everything ;-)

main.cpp
will-v-pi: Done - https://github.com/raspberrypi/pico-sdk/pull/3015

README.md
will-v-pi:
```suggestion
The `bdev` commands are for interacting with block devices in Flash. The block device location can either be determined using binary info, or you can specify a partition to use as a block device. The commands work with a LittleFS filesystem, or a FatFS filesystem.
```

README.md
will-v-pi: Yes, these are applicable to all commands that can force a reboot. Maybe the device selection could be refactored to show cleaner help text as it is the same for all commands (eg a `picotool help device_selection` command to list this help, rather than listing it out for all commands), but that would be for a different PR to do.

main.cpp
will-v-pi: Added - you can now pass the partition number, name or ID (prioritised in that order)

main.cpp
will-v-pi: Added options to override the flags - `--force-formattable` and `--force-writeable`

README.md
```diff
@@ -1703,7 +1703,7 @@ BDEV LS:
 
 SYNOPSIS:
     picotool bdev ls <dirname> [-r] [-p <partition number>] [--partition-name <partition name>] [--partition-id <partition id>]
-                [--filesystem <fs>] [--format] [device-selection]
+                [--filesystem <fs>] [--format] [--force-formattable] [--force-writeable] [device-selection]
```
lurch: Does the `bdev ls` command really need a `--force-writeable` command, given that `ls` will presumably never result in any writes?

will-v-pi: Technically yes, in order to allow formatting a read-only drive - not useful, but technically valid

README.md
will-v-pi: Added

main.cpp
will-v-pi: Fixed

README.md
will-v-pi: Coming back to recursive copying, I think that is trickier, as it needs a recursive walk on the host side - I think I won't add that in this PR, but maybe a future `-r` argument could be added

README.md
lurch: Huh, so merely specifying `--force-formattable` wouldn't allow a read-only drive to be formatted? That might be surprising behaviour? :thinking:

will-v-pi: It throws a reasonable error - I think it's fine, as it requires you to stop and think even more about whether you should format the drive.
```
$ picotool bdev ls --partition-id 0 --format --force-formattable --filesystem littlefs
/
Formatting LittleFS file system
ERROR: This block device is not writeable
```

In this case, if you then pass `--force-writeable` too you still get an error, because the partition is only `NSBOOT(R)`
```
$ picotool bdev ls --partition-id 0 --format --force-formattable --force-writeable --filesystem littlefs
/
Formatting LittleFS file system
ERROR: The RP2350 device returned an error: permission failure
```

will-v-pi:
> might be worth adding some tests to make sure none of the commands delete everything ;-)

`picotool` doesn't have any on-device test infrastructure - also, the commands will delete everything if you tell them to (eg `--format --filesystem fatfs` when the device has an existing `littlefs` filesystem). I think the filesystem libraries are already tested to ensure they don't delete when they're not supposed to.

README.md
lurch: Hmmm, now that there's an explicit `bdev format` command (hooray), _maybe_ it doesn't make sense for the `bdev ls` command to have a `--format` flag, because presumably a freshly-formatted partition will always be empty? :thinking:

will-v-pi: It has uses - you might have a provisioning system that checks for the existence of a specifc file on a board, but which needs to deal with both formatted and un-formatted boards, and is expected to output formatted boards. You could use the same command for all of them because all `bdev` commands come with the ability to format first:
```
picotool bdev ls --format --filesystem littlefs
```

lurch: That seems a bit convoluted, but fair enough :slightly_smiling_face: 
If somebody wanted to do things "the long way round" i.e. avoiding any risk of data-loss, is there a `picotool bdev` command to check if a block device is correctly formatted or not?

lurch:
I wonder if it might be worth pulling out the 1145db7 commit into a separate PR, to minimise the unrelated README churn in **this** PR? :thinking:

main.cpp
```diff
@@ -732,12 +781,14 @@ auto device_selection =
...
 
-#define optional_untyped_file_selection_x(name, i)\
+#define optional_untyped_path_selection_x(name, desc, i)\
 (\
     value(name).with_exclusion_filter([](const string &value) {\
             return value.find_first_of('-') == 0;\
-        }).set(settings.filenames[i]).min(0) % "The file name"\
-).min(0).doc_non_optional(true)
+        }).set(settings.filenames[i]).min(0) % desc\
+).min(0)
```
lurch, line 790: Do the other `opitional_...` macros (e.g. `optional_file_selection_x` and `optional_typed_file_selection_x`) also need the `.doc_non_optional(true)` to be removed?

README.md
will-v-pi: Yes, they can use any command except format, and the command will throw an error if the device wasn’t correctly formatted. If you pass `--filesystem` as well, it will throw an error if the device isn’t formatted to the specified file system.

main.cpp
will-v-pi: No, it’s only this one - the others all have 2 options inside the brackets (file name and type), so seem to need the `doc_non_optional` to be displayed correctly - but this one only has 1 option inside the brackets, so is displayed as optional already. I’m not sure why that’s the case though…

will-v-pi:
> I wonder if it might be worth pulling out the [1145db7](https://github.com/raspberrypi/picotool/commit/1145db7bde8144d280cc3412d22e9b27c2216db0) commit into a separate PR, to minimise the unrelated README churn in **this** PR? 🤔

This PR is getting merged imminently, so I’ll leave it in this PR to avoid creating mess

README.md
lurch: Cool, thanks. Is that behaviour worth documenting (i.e. how to check if a block device is formatted, without writing to it), or do you think that it's sufficiently obvious that people will figure it out for themselves?

main.cpp
lurch: Fair enough - I guess this is one of those "if it works, don't touch it" things :rofl:

README.md
will-v-pi: I think it's clear enough?
```
$ picotool bdev ls
ERROR: No file system detected - to format the drive use `--format --filesystem <littlefs|fatfs>`

$ picotool bdev ls --filesystem littlefs
/
ERROR: LittleFS file system is corrupted - add --format flag to format it (this may result in data loss)

$ picotool bdev ls --filesystem fatfs
/
ERROR: FatFS file system is corrupted - add --format flag to format it (this may result in data loss)
```

will-v-pi:
> might be worth adding some tests to make sure none of the commands delete everything ;-)

I have added a `bdev_test.sh` script which runs some quick tests to check `rm` only removes the file it's directed to - this script requires having an RP2350 device, so must be run locally to test

main.cpp
```diff
@@ -4354,35 +4579,66 @@ std::shared_ptr<vector<tuple<uint32_t, uint32_t>>> get_partitions(picoboot::conn
...
+            return nullptr;
         }
     }
 
-    return std::make_shared<vector<tuple<uint32_t, uint32_t>>>(ret);
+    return std::make_shared<tuple<resident_partition_t, std::shared_ptr<vector<partition_details>>>>(std::make_tuple(unpartitioned, std::make_shared<vector<partition_details>>(ret)));
+}
+
+std::shared_ptr<vector<partition_details>> get_partitions(picoboot::connection &con) {
+    return std::get<1>(*get_partition_info(con));
```
aswild, line 4641: this operation crashes `picotool info` on RP2040 targets that don't support partition info. When `model->supports_partition_table()` returns `false`, as it does on RP2040, `get_partition_info()` returns `nullptr` and then this deref fails (either with a failed assert in debug mode, or a segfault in release mode).

This code needs to forward on `nullptr` (everywhere that calls `get_partitions` already makes sure its result is non-null before using it)

```diff
diff --git a/main.cpp b/main.cpp
index 0c3b093..b2d37e4 100644
--- a/main.cpp
+++ b/main.cpp
@@ -4670,7 +4670,11 @@ std::shared_ptr<tuple<resident_partition_t, std::shared_ptr<vector<partition_det
 }

 std::shared_ptr<vector<partition_details>> get_partitions(picoboot::connection &con) {
-    return std::get<1>(*get_partition_info(con));
+    auto info = get_partition_info(con);
+    if (info) {
+        return std::get<1>(*info);
+    }
+    return nullptr;
 }
 #endif

```

lurch: > this operation crashes `picotool info` on RP2040 targets that don't support partition info.

Thanks for the report. Could you create a new issue please?

PR #196, opened 2025-01-28
Fix typo for Arm-NS family name parsing

Wren6991:
We check for the Arm-S name twice, so this branch will never be hit


will-v-pi (approved):
Thanks - looks like the rest of the file was handling it correctly, just this one was wrong

PR #203, opened 2025-02-17
Speedup coprodis

### LINKED ISSUE #106: Picotool coprodis unusably slow with large program (W11)
MMBasic creates an image which is just under a Megabyte (rp2040). elf2uf2 used to create the uf2 in a couple of seconds. picotool takes several minutes and a processor is running a maximum during this period (W11 PC with I7-12700). It does eventually complete and appears to create a valid uf2.
I believe the problem is in the creation of the dis file which is 11.4Mbytes. If I kill picotool using the task manager the UF2 is created but the dis creation fails

kilograham:
fixes #106

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -1237,6 +1237,8 @@ struct coprodis_command : public cmd {
     string get_doc() const override {
         return "Post-process coprocessor instructions in disassembly files.";
     }
+
+    bool decode_line(uint32_t val, char *buf, size_t buf_len);
 };
 
 struct help_command : public cmd {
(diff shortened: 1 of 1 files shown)


main.cpp
```diff
@@ -6319,673 +6330,698 @@ string cpu_reg(int val) {
...
+        MRC = 0x04,
+        MRC2 = 0x08,
+        MCRR = 0x10,
+        MCRR2 = 0x20,
+        MRRC = 0x40,
+        MRRC2 = 0x80,
+        CDP = 0x100,
+    } type = NONE;
+    uint32_t mccrbits = (val & 0xfff00000) >> 20;
+    uint32_t cdpbits = val & 0xff000010;
```
lurch, line 6363: Hmm... `mcrbits = val & 0xff100010` and `cdpbits = val & 0xff000010` so it looks like `cdpbits` is actually a **subset** of `mcrbits`?

If the intention is that `val` only ever decodes to _one_ of the `type` enums, perhaps there ought to be a `if (type != NONE) {` around the `switch (mccrbits) {` and `switch (cdpbits) {` blocks?

(And I guess you could left-shift the `case` values in the `switch (mccrbits)` by 20 bits, and then you would't need the `>> 20` when masking off `mccrbits`?)

main.cpp
```diff
@@ -6319,673 +6330,698 @@ string cpu_reg(int val) {
...
+bool coprodis_command::execute(device_map &devices) {
+    auto in = get_file(ios::in);
+    std::stringstream buffer;
+    buffer << in->rdbuf();
 
-    if (proc_insts.size() == 0) {
-        *out << buffer.str();
-        return false;
-    }
+    std::regex instruction(
```
lurch, line 6944: Is this now unused? (and if so should it be deleted, to stop it being a red-herring?)

main.cpp
```diff
@@ -6319,673 +6330,698 @@ string cpu_reg(int val) {
...
-                mcr = true;
-                inst = "mrc";
-                break;
-            case 0xfe100010:
-                mcr = true;
-                inst = "mrc2";
-                break;
-            default:
-                break;
+static bool parse_hex(const std::string& value, uint32_t& out) {
```
lurch, line 6333: Is this `parse_hex` function unused?

PR #206, opened 2025-02-24
Speedup some picotool commands

### LINKED ISSUE #204: Performance with large binaries still could be improved
This is a regression, presumably when adding Metadata dumping...

`picotool info -a` can take 3 or 4 seconds for a 2M binary

will-v-pi:
This offers substantial performance improvements for `info` (with no signing/hashing) and `save`/`verify`, and small improvements for `load`. The performance figures for a 4MiB binary on Pico 2 are below, and on Pico the figures are also improved (although with less of an improvement for `info` due to the Pico not including metadata blocks). This fixes #204.

| Command | speedups | 2.1.1 | 2.1.0 | 2.0.0 |
|--------|--------|--------|--------|--------|
| `info` | 0.056 | 9.156 | 18.488 | 9.346 |
| `info` (hashed binary) | 9.156 | 9.173 | 36.747 | 18.408 |
| `save` | 9.8 | 21.747 | 31.28 | 35.869 |
| `save --range` | 9.744 | 40.058 | 23.325 | 26.087 |
| `load` | 65.121 | 70.688 | 72.645 | 89.636 |

The main fixes have been:
- Batching reads/writes into >4096 byte chunks, rather than the existing smaller chunk sizes (eg 256 byte chunks for `save`)
- Doing a sparse read of the block loop when not verifying the binary hash and only read the whole binary when verifying the hash, rather than always reading the whole binary in case the hash needs to be verified.

Also includes 3 drive-by fixes:
- Fixed a bug with truncated binaries, where the metadata search would get into an infinite loop - this fix was already in `get_last_block` but was missing from `get_all_blocks`
- Replace `<unknown>` printout for `PICOBOOT_UNKNOWN_ERROR` with `unknown error`, to clarify the error was not an unknown error value, just an unknown error - the `<unknown>` printout is still used for other unknown error values
- Workaround picoboot interface not supporting >4MiB reads - this only crops up when validating the hash/signature for >4MiB binaries, as it reads the whole binary in one go

### DIFF
--- a/bintool/bintool.cpp
+++ b/bintool/bintool.cpp
@@ -315,52 +315,6 @@ block place_new_block(elf_file *elf, std::unique_ptr<block> &first_block) {
 }
 
 
-std::unique_ptr<block> get_last_block(std::vector<uint8_t> &bin, uint32_t storage_addr, std::unique_ptr<block> &first_block, get_more_bin_cb more_cb) {
-    uint32_t next_block_addr = first_block->physical_addr + first_block->next_block_rel;
-    std::unique_ptr<block> new_first_block;
-    uint32_t read_size = PICOBIN_MAX_BLOCK_SIZE;
-    uint32_t current_bin_start = storage_addr;
-    while (true) {
-        if (next_block_addr + read_size > current_bin_start + bin.size() && more_cb != nullptr) {
-            DEBUG_LOG("Reading into bin %08x+%x\n", next_block_addr, read_size);
-            more_cb(bin, next_block_addr, read_size);
-            current_bin_start = next_block_addr;
-        }
-        auto offset = next_block_addr - current_bin_start;
-        std::vector<uint32_t> words = lsb_bytes_to_words(bin.begin() + offset, bin.end());
-        if (words.front() != PICOBIN_BLOCK_MARKER_START) {
-            fail(ERROR_UNKNOWN, "Block loop is not valid - no block found at %08x\n", (int)(next_block_addr));
-        }
-        words.erase(words.begin());
-        DEBUG_LOG("Checking block at %x\n", next_block_addr);
-        DEBUG_LOG("Starts with %x %x %x %x\n", words[0], words[1], words[2], words[3]);
-        for(auto next_item = words.begin(); next_item < words.end(); ) {
-            unsigned int size = item::decode_size(*next_item);
-            if ((uint8_t)*next_item == PICOBIN_BLOCK_ITEM_2BS_LAST) {
-                if (size == next_item - words.begin()) {
-                    if (next_item < words.end() && next_item[2] == PICOBIN_BLOCK_MARKER_END) {
-                        DEBUG_LOG("is a valid block\n");
-                        new_first_block = block::parse(next_block_addr, next_item + 1, words.begin(), words.begin() + size);
-                        break;
-                    }
-                }
-            } else {
-                next_item += size;
-            }
-        }
-        if (new_first_block->physical_addr + new_first_block->next_block_rel == first_block->physical_addr) {
-            DEBUG_LOG("Found last block in block loop\n");
-            break;
-        } else {
-            DEBUG_LOG("Continue looping\n");
-            next_block_addr = new_first_block->physical_addr + new_first_block->next_block_rel;
-            new_first_block.reset();
-        }
-    }
-    return new_first_block;
-}
-
-
 std::vector<std::unique_ptr<block>> get_all_blocks(std::vector<uint8_t> &bin, uint32_t storage_addr, std::unique_ptr<block> &first_block, get_more_bin_cb more_cb) {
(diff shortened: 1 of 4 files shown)


main.cpp
```diff
@@ -1867,7 +1868,7 @@ struct picoboot_memory_access : public memory_access {
     }
 
     void read(uint32_t address, uint8_t *buffer, unsigned int size, __unused bool zero_fill) override {
-        if (flash == get_memory_type(address, model)) {
+        if (flash == get_memory_type(address, model) && settings.use_flash_cache) {
```
lurch: Is it worth swapping these conditionals around, as I guess if `settings.use_flash_cache` returns false then there's no need to check `flash == get_memory_type(address, model)` ?

bintool/bintool.cpp
```diff
@@ -362,12 +365,15 @@ std::vector<std::unique_ptr<block>> get_all_blocks(std::vector<uint8_t> &bin, ui
     uint32_t next_block_addr = first_block->physical_addr + first_block->next_block_rel;
     std::vector<std::unique_ptr<block>> all_blocks;
     uint32_t read_size = PICOBIN_MAX_BLOCK_SIZE;
+    uint32_t current_bin_start = storage_addr;
     while (true) {
         std::unique_ptr<block> new_first_block;
-        auto offset = next_block_addr - storage_addr;
-        if (offset + read_size > bin.size() && more_cb != nullptr) {
-            more_cb(bin, offset + read_size);
+        if (next_block_addr + read_size > current_bin_start + bin.size() && more_cb != nullptr) {
```
lurch, line 325: The change here (lines 368-376) looks identical to the previous change (on lines 322-329). Does that suggest that there's some common code that might be worth pulling out into a separate function? :shrug:

will-v-pi: They had entirely common code, so I've changed `get_last_block` to now call `get_all_blocks` and just return the last block.

bintool/bintool.cpp
```diff
@@ -315,60 +315,23 @@ block place_new_block(elf_file *elf, std::unique_ptr<block> &first_block) {
...
```
kilograham, line 325: seems like we should track which blocks we've already seen to catch loops or is that done elsewhere?

main.cpp
```diff
@@ -4303,6 +4308,7 @@ bool save_command::execute(device_map &devices) {
         fail(ERROR_NOT_POSSIBLE, "Save range crosses unmapped memory");
     }
     uint32_t size = end - start;
+    uint32_t chunk_size = std::max(size / 100, FLASH_SECTOR_ERASE_SIZE);
```
kilograham: I think i'd prefer chunk_size / 100 rounded up to the next FLASH_SECTOR_ERASE_SIZE multiple

kilograham: not critical for read, but perhaps tidier and more consistent with others

kilograham (changes requested):

main.cpp
```diff
@@ -4592,7 +4602,9 @@ bool load_guts(picoboot::connection con, iostream_memory_access &file_access) {
         // new scope for progress bar
         {
             progress_bar bar("Loading into " + memory_names[type] + ": ");
-            uint32_t batch_size = FLASH_SECTOR_ERASE_SIZE;
+            // Use batches of size/100 rounded up to FLASH_SECTOR_ERASE_SIZE
+            uint32_t batch_size = std::max(FLASH_SECTOR_ERASE_SIZE,
```
kilograham: arguably the std::max is unnecessary, since everything but 0 would be rounded up anyway

main.cpp
```diff
@@ -4637,7 +4649,7 @@ bool load_guts(picoboot::connection con, iostream_memory_access &file_access) {
             bool ok = true;
             {
                 progress_bar bar("Verifying " + memory_names[type] + ": ");
-                uint32_t batch_size = FLASH_SECTOR_ERASE_SIZE;
+                uint32_t batch_size = std::max(mem_range.len() / 100, FLASH_SECTOR_ERASE_SIZE);
```
kilograham: ditto batch size - perhaps make a function

main.cpp
```diff
@@ -5272,7 +5284,7 @@ bool verify_command::execute(device_map &devices) {
                     progress_bar bar("Verifying " + memory_names[t1] + ": ");
                     vector<uint8_t> file_buf;
                     vector<uint8_t> device_buf;
-                    uint32_t batch_size = 1024;
+                    uint32_t batch_size = std::max(mem_range.len() / 100, FLASH_SECTOR_ERASE_SIZE);
```
kilograham: ditto

bintool/bintool.cpp
will-v-pi: Valid block loops are caught below - `Found last block in block loop`

Lollipop block loops are not caught so will (probably) loop forever, but they are invalid anyway, and will trigger E22

kilograham (approved):

PR #207, opened 2025-02-24
Add support for creating self-decrypting binaries, and use 4-way AES key shares instead of just the AES key

### LINKED ISSUE #210: "ERROR: Found overlapping memory" on encrypted binary
If I run:
`picotool info -all ../build/hello_serial_enc.uf2`

I am getting following error:

```
File ../build/hello_serial_enc.uf2 family ID 'rp2350-arm-s':

ERROR: Found overlapping memory ranges 0x20000000->0x20003d5c and 0x20000000->20000014
```

I tried that with hello_serial_enc from pico examples and also with my own encrypted binary

### LINKED ISSUE #235: add_custom_command uses wrong arguments
e.g.

```
    add_custom_command(TARGET xip_ram_perms
        COMMAND ${CMAKE_COMMAND} -E copy ${XIP_RAM_PERMS_ELF} ${CMAKE_BINARY_DIR}/xip_ram_perms.elf
        DEPENDS xip_ram_perms
    )
```

Newest CMake now complains about this:

1. you must now specify `POST_BUILD` explicitly
2. `DEPENDS` was always incorrect I guess

will-v-pi:
This adds support for creating self-decrypting binaries, which can be created using
```bash
picotool encrypt --sign --embed hello_world.elf hello_world.enc.elf my_aesfile.bin my_iv_salt.bin my_sigfile.pem my_otp.json
```

This introduces a breaking change to `picotool encrypt` to take a 4-way AES key share, rather than just taking an AES key, as this makes it simpler to mask the power signature when decrypting the binary. The only difference from a user perspective is that they now need to use a 1024 bit binary file instead of the 256 bit file used before. The AES key is derived from the 4-way share as follows:
```c
aes_key.words[i] = aes_key_share.words[i*4]
                 ^ aes_key_share.words[i*4 + 1]
                 ^ aes_key_share.words[i*4 + 2]
                 ^ aes_key_share.words[i*4 + 3];
```

This also introduces a breaking change that `picotool encrypt` now requires an IV salt binary to be passed to it as the 4th file, so the signing_key is now the 5th file and the OTP JSON is now the 6th file.

This PR should be merged in parallel with https://github.com/raspberrypi/pico-sdk/pull/2315 and https://github.com/raspberrypi/pico-examples/pull/619 to avoid breaking the pico-examples encrypted bootloader compilation


README.md
```diff
@@ -678,7 +678,8 @@ The encrypted binary will have the following structure:
 - Padding to ensure the encrypted length is a multiple of 4 words
 - Signature metadata block
 
-The AES key must be provided as a .bin file of the 256 bit AES key to be used for encryption.
+The AES key must be provided as a .bin file containing a 4-way share of the 256 bit AES key to be used for encryption.
```
lurch: Is there an (e.g.) `openssl` command that can be used to create this "4-way share of the 256 bit AES key" ?

will-v-pi: It’s just a 1024 bit random number, rather than a 256 bit one - see https://github.com/raspberrypi/pico-examples/pull/619 for generation suggestions

bintool/mbedtls_wrapper.c
```diff
@@ -40,15 +40,62 @@ void mb_sha256_buffer(const uint8_t *data, size_t len, message_digest_t *digest_
...
+}
+#endif
+
+void mb_aes256_buffer(const uint8_t *data, size_t len, uint8_t *data_out, const aes_key_t *key, iv_t *iv) {
     mbedtls_aes_context aes;
 
     assert(len % 16 == 0);
 
     mbedtls_aes_setkey_enc(&aes, key->bytes, 256);
+    uint8_t xor_working_block[16] = {0};
```
lurch: Looks like `uint8_t xor_working_block[16]` could be moved inside the `#if IV0_XOR` ?

bintool/mbedtls_wrapper.c
```diff
@@ -40,15 +40,62 @@ void mb_sha256_buffer(const uint8_t *data, size_t len, message_digest_t *digest_
...
+    size_t n = 0;
+    uint32_t counter = 0;
+
+    assert(length == (uint32_t)length);
+
+    while (length--) {
+        if (n == 0) {
+            for (int i = 16; i > 0; i--) {
+                nonce_xor[i-1] = iv0[i-1];
+                if (i - (int)(16 - sizeof(counter)) > (int)0) {
```
lurch: I _think_ that `if (i - (int)(16 - sizeof(counter)) > (int)0)` is basically the same as `if (i > 12)` ?
Is it being done "the long way round" to aid with the "power masking"? (And if so, is there a risk that a compiler might be able to optimise it to just `if (i > 12)`?)

README.md
lurch: Ahhh, I'd misinterpreted this to mean that the 1024-bit keyshare was directly generated from the 256-bit key; rather than the 256-bit key being generated from the (random) 1024-bit keyshare :upside_down_face:  :joy:

README.md
```diff
@@ -678,7 +678,8 @@ The encrypted binary will have the following structure:
...
+Eg if the 256 bit key is X, the bin file should contain the 256 bit numbers A B C D where X = A ^ B ^ C ^ D.
```
lurch: This is the line that confused me (it implies that you know X, and then use that to generate A B C D). Perhaps it should be reworded slightly?
"E.g. if the bin file contains the 256-bit numbers A B C D, the 256-bit key X is calculated as X = A ^ B ^ C ^ D." :man_shrugging:  (Although my terminology may be wrong, as this is the first time I've encountered "key shares")

bintool/mbedtls_wrapper.c
will-v-pi: No power masking (we don't care about that in picotool, only when running code on the device), just my poor code

README.md
will-v-pi: You can do it either way round - if you want a specific key X then you can generate A, B, C & D to match your X, although have to be very careful to make sure they are properly random. The normal use case would be you don't need a specific key, so just randomly generate A, B, C & D and use whatever X that gives, so rewording is probably sensible

will-v-pi: Actually, I've reworded it to reflect that we're using word-wise shares, so probably needs a re-read to check it still makes sense...

CMakeLists.txt
```diff
@@ -57,11 +57,70 @@ endif()
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake)
 
+# To configure mbedtls
+# todo make the configuration better
+set(MBEDTLS_CONFIG_FILE "mbedtls_config.h")
+add_compile_options(-I${CMAKE_SOURCE_DIR}/lib/include)
```
lurch: Would it be better to use https://cmake.org/cmake/help/latest/command/include_directories.html rather than `add_compile_options` ?

enc_bootloader.h
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#pragma once
+
+#include <memory>
+#include <fstream>
+
+std::shared_ptr<std::iostream> get_enc_bootloader(bool use_mbedtls = false);
```
lurch: This function doesn't seem to be used anywhere? :confused:

will-v-pi: It is, in `encrypt_command` in main.cpp

lurch: DOH! Github was showing me the "Large diffs are not rendered by default." for `main.cpp` when I initially tried searching for where this function was used :man_facepalming: 

_However_ I wonder if it might be worth renaming `enc_bootloader.cpp` and `enc_bootloader.h` to something else, in order to avoid any possible confusion with the `enc_bootloader.c` that's also in this PR? :thinking:  (AFAICT the former is code that gets run on your PC, and the latter is code that gets run on your Pico?)

will-v-pi: I've renamed `enc_bootloader.cpp/.h` to `get_enc_bootloader.cpp/.h`, and done the same for `xip_ram_perms.cpp`, as that's more what the cpp files are doing (they just locate the ELFs to run on the Pico)

CMakeLists.txt
will-v-pi: Yep - have changed it to that

CMakeLists.txt
```diff
@@ -57,11 +57,70 @@ endif()
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake)
 
+# To configure mbedtls
```
kilograham: seems like these are the same, and really i guess the TODO should be in the header

CMakeLists.txt
```diff
@@ -57,11 +57,70 @@ endif()
...
+# todo make the configuration better
+set(MBEDTLS_CONFIG_FILE "mbedtls_config.h")
+include_directories(${CMAKE_SOURCE_DIR}/lib/include)
```
kilograham: yuk`include_directories`

CMakeLists.txt
```diff
@@ -169,6 +228,15 @@ if (NOT PICOTOOL_NO_LIBUSB)
     endif()
 endif()
 
+if (TARGET mbedtls)
+    add_custom_target(binary_data_no_libusb DEPENDS
```
kilograham: `binary_data` is confusing because it has meaning in SDK; maybe something with "embedded"

README.md
```diff
@@ -678,7 +678,9 @@ The encrypted binary will have the following structure:
 - Padding to ensure the encrypted length is a multiple of 4 words
 - Signature metadata block
 
-The AES key must be provided as a .bin file of the 256 bit AES key to be used for encryption.
+The AES key must be provided as a 1024 bit .bin file containing 4-way shares of each word of the 256 bit AES key to be used for encryption.
```
kilograham: I'm very confused as a user what I'm actually supposed to do. should i pass a file of length 128?

bintool/bintool.cpp
```diff
@@ -648,7 +648,7 @@ std::vector<uint8_t> get_lm_hash_data(elf_file *elf, block *new_block, bool clea
             if (data.size() != seg->physical_size()) {
                 fail(ERROR_INCOMPATIBLE, "Elf segment physical size (%" PRIx32 ") does not match data size in file (%zx)", seg->physical_size(), data.size());
             }
-            if (seg->physical_size() && seg->physical_address() < new_block->physical_addr) {
+            if (seg->physical_size()) {
```
kilograham, line 673: what was this change?

enc_bootloader/CMakeLists.txt
```diff
@@ -0,0 +1,104 @@
...
+    unset(ENV{LDFLAGS})
+
+    # Pull in SDK (must be before project)
+    include(${PICO_SDK_PATH}/external/pico_sdk_import.cmake)
+
+    project(enc_bootloader C CXX ASM)
+    set(CMAKE_C_STANDARD 11)
+    set(CMAKE_CXX_STANDARD 17)
+
+    if (PICO_SDK_VERSION_STRING VERSION_LESS "2.0.0")
```
kilograham: Presumably needs 2.1.2

enc_bootloader/enc_bootloader.c
```diff
@@ -0,0 +1,202 @@
...
+
+// The function lock_key() is called from decrypt() after key initialisation is complete and before decryption begins.
+// That is a suitable point to lock the OTP area where key information is stored.
+void lock_key() {
+    otp_hw->sw_lock[otp_key_page] = 0xf;
+}
+
+
+int main() {
+    // This works around E21 by locking down the page with a key,
```
kilograham: not sure what the point of this is - if using a well known key. Does this really help anyway? E21 or not, there are other errata that make it possible to run unsigned code, which would let them legitimately read OTP (unless it is page locked like this, but then where do we keep the page key?)

kilograham:
1. I had forgotten? that `encrypt` was already an available option. Was that used with the pre-existing `pico-example`? How does this change affect that example (is the `encrypt` command now backwards compatible? (edit: i guess i should read the PR description!)
2. I worry about rollback versions (and indeed other versions) - we should probably copy that across to the decrypting header
3. I believe the dealing with TBYB and friends IS handled (due to pico-sdk PR) just wanted to double check with you - i still need to delve into this PR more deeply

will-v-pi:
> 1. I had forgotten? that `encrypt` was already an available option. Was that used with the pre-existing `pico-example`? How does this change affect that example (is the `encrypt` command now backwards compatible? (edit: i guess i should read the PR description!)
> 2. I worry about rollback versions (and indeed other versions) - we should probably copy that across to the decrypting header
> 3. I believe the dealing with TBYB and friends IS handled (due to pico-sdk PR) just wanted to double check with you - i still need to delve into this PR more deeply

1. I've modified the example to use the new command https://github.com/raspberrypi/pico-examples/pull/619 - but yes, the SDK CMake function and the picotool command are not backwards compatible.
2. They should have been copied across, along with TBYB - lines 5124-5139 in main.cpp - but this wasn't working as the block choice code would choose the first block to copy data from. I've changed this so it will now choose to copy the last block if it has an IMAGE_DEF, otherwise choose the first block, as the last block will be the better choice.
3. For self-decrypting binaries, TBYB and everything else is handled entirely by the bootrom - the decryption stage shouldn't time out the watchdog, so the user binary can call `explicit_buy` as usual. The decryption stage just gets loaded into SRAM along with the encrypted binary by the bootrom, and decrypts it in place, so doesn't need any handling for TBYB etc. For encrypted binaries with a separate decrypting bootloader, the handling is improved by the `rom_pick_ab_update_partition` PR.

enc_bootloader/enc_bootloader.c
will-v-pi: This was originally just added to the encrypted bootloader example as a quick demo, and I thought I'd keep it in as it worked around E21 in a simple way. I can remove it if you'd prefer.

bintool/bintool.cpp
will-v-pi: https://github.com/raspberrypi/picotool/commit/011111d8dfe291d66531cd82a4704df8688b7f54 - it allows signing code in scratch

README.md
will-v-pi: You need to pass a file of length 1024 bits, or 128 bytes - I went with bits, because you normally talk about key length in bits rather than bytes

CMakeLists.txt
will-v-pi: @lurch suggested it https://github.com/raspberrypi/picotool/pull/207#discussion_r2003612398 - I can go back to what it was before if you prefer

It had to be a bit yuck, as it needs to be applied to the mbedtls build without modifying it's CMake files

kilograham: oh ok. maybe a comment then

enc_bootloader/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ if (NOT USE_PRECOMPILED)
     set(CMAKE_C_STANDARD 11)
     set(CMAKE_CXX_STANDARD 17)
 
-    if (PICO_SDK_VERSION_STRING VERSION_LESS "2.0.0")
+    if (PICO_SDK_VERSION_STRING VERSION_LESS "2.1.2")
         message(FATAL_ERROR "Raspberry Pi Pico SDK version 2.0.0 (or later) required. Your version is ${PICO_SDK_VERSION_STRING}")
```
kilograham: was hoping you'd change the message too ;-)

lib/include/mbedtls_config.h
```diff
@@ -15,6 +15,8 @@
 #ifndef MBEDTLS_CONFIG_H
 #define MBEDTLS_CONFIG_H
 
+// todo make the configuration better
```
kilograham: maybe "// todo fine tune this config file"

enc_bootloader/CMakeLists.txt
will-v-pi: oops - thanks

kilograham:
A few other random thoughts

1. I think we should support passing a 256 bit AES key and generating 4 shares (I dont think this tells you much if the AES key is secure random even if picotool does less well -though it should be using dev/random already, right via c++ lib if i reacll)
2. I think we should support passing the AES key and the IV salt as a hex on the command line too. could probably live with just treating anything starting with 0x as hex not filename, and check it is the right length. 
3. Can we lose the `-t` on the AES key and salt unless there are useful different file types

enc_bootloader/enc_bootloader.c
kilograham: yeah, i mean it burns a user key (right?) for no particular benefit

kilograham:
I wonder if we should explicitly zero unused RAM in the LOAD_MAP of the self-encrypting binary (probably)

README.md
lurch: Might be worth mentioning the length in both bits **and** byes? Since as William points out key-lengths are measured in bits, but the `ls` command only shows file-sizes in bytes.

enc_bootloader/enc_bootloader.c
will-v-pi: Have removed it, and removed the corresponding OTP writing in the examples

README.md
```diff
@@ -678,9 +678,11 @@ The encrypted binary will have the following structure:
 - Padding to ensure the encrypted length is a multiple of 4 words
 - Signature metadata block
 
-The AES key must be provided as a 1024 bit (128 byte) .bin file containing 4-way shares of each word of the 256 bit AES key to be used for encryption.
-This should usually be generated by creating a random file of length 128 bytes and deriving the AES key from that, rather than crafting a file to match a specific AES key, to ensure the shares are random.
+The AES key can either be provided as a 1024 bit (128 byte) .bin file containing 4-way shares of each word of the 256 bit AES key to be used for encryption, or as a 256 bit (32 byte) .bin file containing the 256 bit AES key.
+The key share should usually be generated by creating a random file of length 128 bytes and deriving the AES key from that, rather than crafting a file to match a specific AES key, to ensure the shares are random.
+When passing a 256 bit AES key (as opposed to a 1028 bit 4-way AES key share), picotool will generate a random key share to match that key, by generating 3 random shares (A, B & C) and calculating a 4th share (D) to match the provided key.
```
lurch: IMHO this needs rewording for clarity, as it's currently pretty confusing with the two different methods intermingled. (Also, you've written "1028 bit" which I'm assuming is wrong :wink: )

main.cpp
```diff
@@ -4999,10 +5000,29 @@ bool encrypt_command::execute(device_map &devices) {
...
+        std::random_device rand{};
+        assert(rand.max() - rand.min() >= 256);
+        for(int i=0; i < 8; i++) {
+            for (int j=0; j < 12; j++) {
+                aes_key_share.bytes[i*16 + j] = rand();
+            }
+            aes_key_share.words[i*4 + 3] = tmp_key.words[i]
+                                        ^ aes_key_share.words[i*4 + 2]
+                                        ^ aes_key_share.words[i*4 + 1]
+                                        ^ aes_key_share.words[i*4];
```
lurch: It probably doesn't matter, but this appears to be doing `D[i] = X[i] ^ C[i] ^ B[i] ^ A[i]` which is different to the `D[i] = X[i] ^ A[i] ^ B[i] ^ C[i]` described in the README?

will-v-pi: Yeah, it doesn't matter, because XOR is associative and commutative, so can be done in any order

main.cpp
```diff
@@ -4999,10 +5000,29 @@ bool encrypt_command::execute(device_map &devices) {
     aes_key_share_t aes_key_share;
     aes_file->seekg(0, std::ios::end);
     if (aes_file->tellg() != 128) {
-        fail(ERROR_INCOMPATIBLE, "The AES key share must be a 128 byte file (the supplied file is %d bytes)", aes_file->tellg());
+        // Generate a random key share from 256-bit key
+        if (aes_file->tellg() != 32) {
+            fail(ERROR_INCOMPATIBLE, "The AES key share must be a 128 byte key share, or a 32 byte key (the supplied file is %d bytes)", aes_file->tellg());
```
lurch: Perhaps:
```c++
streampos aes_key_size = aes_file->tellg();
if (aes_key_size == 32) {
    // Generate a random key share from 256-bit key
} else if (aes_key_size == 128) {
    // read the share directly
} else {
    fail(ERROR_INCOMPATIBLE, "The AES key share must be a 128 byte key share, or a 32 byte key (the supplied file is %d bytes)", aes_key_size);
}
```
is easier to read than:
```c++
if (aes_file->tellg() != 128) {
    if (aes_file->tellg() != 32) {
        fail(ERROR_INCOMPATIBLE, "The AES key share must be a 128 byte key share, or a 32 byte key (the supplied file is %d bytes)", aes_file->tellg());
    }
    // Generate a random key share from 256-bit key
} else {
    // read the share directly
}

**Comment by will-v-pi** [comment]
> 1. I think we should support passing a 256 bit AES key and generating 4 shares (I dont think this tells you much if the AES key is secure random even if picotool does less well -though it should be using dev/random already, right via c++ lib if i reacll)
> 2. I think we should support passing the AES key and the IV salt as a hex on the command line too. could probably live with just treating anything starting with 0x as hex not filename, and check it is the right length.
> 3. Can we lose the `-t` on the AES key and salt unless there are useful different file types

I've done these, and for 3. have also removed it from all the `pem` and `json` files which only support one filetype, as those should also always have the correct extension

File: main.cpp (hunk 17 above)
Comment (will-v-pi) [reply main.cpp]: I've fixed the order in the c++ anyway

File: main.cpp [hunk 19]
```diff
@@ -4991,11 +5001,37 @@ bool encrypt_command::execute(device_map &devices) {
         fail(ERROR_ARGS, "Can only sign to same file type");
     }
 
-    if (get_file_type_idx(2) != filetype::bin) {
+    if (!settings.filenames[2].empty() && settings.filenames[2].find("0x") == 0) {
+        // Hex string instead of file
```
lurch: Feels like this "parse a long hex string into an array of byes" probably ought to be abstracted out to a separate function, to avoid code-repetition?

main.cpp
```diff
@@ -4991,11 +5001,37 @@ bool encrypt_command::execute(device_map &devices) {
...
+        }
+    } else if (get_file_type_idx(2) != filetype::bin) {
         fail(ERROR_ARGS, "Can only read AES key share from BIN file");
     }
 
-    if (get_file_type_idx(3) != filetype::bin) {
+    if (!settings.filenames[3].empty() && settings.filenames[3].find("0x") == 0) {
+        // Hex string instead of file
+        if (settings.filenames[3].size() != 34) {
+            fail(ERROR_ARGS, "IV OTP salt hex string must be 16 bytes long");
```
lurch: Does the README need to be updated to mention that the salt can also be passed as a hexadecimal string?

README.md
```diff
@@ -678,7 +678,13 @@ The encrypted binary will have the following structure:
...
 - Signature metadata block
 
-The AES key must be provided as a .bin file of the 256 bit AES key to be used for encryption.
+The AES key can either be provided as a 1024 bit (128 byte) .bin file containing 4-way shares of each word of the 256 bit AES key to be used for encryption, as a 256 bit (32 byte) .bin file containing the 256 bit AES key, or a 256 bit hex string of the AES key (eg 0x0123456789abcdeffedcba98765432100123456789abcdeffedcba9876543210).
+
+The key share should usually be generated by creating a random file of length 128 bytes and deriving the AES key from that, rather than crafting a file to match a specific AES key, to ensure the shares are random.
+However, if you wish to use a specific 256 bit AES key, then picotool will generate a random key share to match that key, by generating 3 random shares (A, B & C) and calculating a 4th share (D) to match the provided key.
+
+With the words stored in the file as A[0], B[0], C[0], D[0], A[1], B[1], etc, C[7], D[7], word i of the key X is X[i] = A[i] ^ B[i] ^ C[i] ^ D[i].
+When generating a key share from a key, picotool calculates it as D[i] = X[i] ^ A[i] ^ B[i] ^ C[i].
```
lurch: Better than it was, but I still think there's too much information being front-loaded in the first sentence (and there are parts which seem no longer applicable). How about _something_ like:

The AES key can either be provided as a 1024-bit (128 byte) key share or as a 256-bit (32 byte) AES key. In the latter case, a key share will be generated for you.

The 4-way key share should be generated by creating a random `.bin` file of length 128 bytes and the AES key will be derived from that. With the words stored in the file as A[0], B[0], C[0], D[0], A[1], B[1], etc., C[7], D[7], word i of the key X is X[i] = A[i] ^ B[i] ^ C[i] ^ D[i].

Alternatively an AES key can be provided directly, as either a 32 byte `.bin` file or as a string of 64 hexadecimal characters (eg 0x0123456789abcdeffedcba98765432100123456789abcdeffedcba9876543210). Picotool will generate a random key share for that key, by generating 3 random shares (A, B & C) and calculating a 4th share (D) to match the provided key (X), with each word i of the key share calculated as D[i] = X[i] ^ A[i] ^ B[i] ^ C[i].

main.cpp
will-v-pi: Yeah, the README needs to be updated to mention the salt

README.md
```diff
@@ -678,28 +678,46 @@ The encrypted binary will have the following structure:
...
 
-With the words stored in the file as A[0], B[0], C[0], D[0], A[1], B[1], etc, C[7], D[7], word i of the key X is X[i] = A[i] ^ B[i] ^ C[i] ^ D[i].
-When generating a key share from a key, picotool calculates it as D[i] = X[i] ^ A[i] ^ B[i] ^ C[i].
+Alternatively an AES key can be provided directly, as either a 32 byte .bin file or as a string of 64 hexadecimal characters (eg 0x0123456789abcdeffedcba98765432100123456789abcdeffedcba9876543210). Picotool will generate a random key share for that key, by generating 3 random shares (A, B & C) and calculating a 4th share (D) to match the provided key (X), with each word i of the 4th share calculated as D[i] = X[i] ^ A[i] ^ B[i] ^ C[i].
+
+The encryption also requires a per-device IV salt in OTP, which should be a 128-bit (16 byte) random file - this can be provided as either a 16 byte .bin file, or a string of 32 hexadecimal characters, similar to the AES key. This per-device salt will be XORed with the IV stored in the binary, to give the IV used by the decryption code.
```
lurch: "128-bit (16 byte) random file" -> "128-bit (16 byte) random value" ?
(since otherwise the bit about the hex-string sounds weird)

lurch: Is it worth clarifying what "IV" stands for? Or does everyone who knows what an "AES key" is, will also know what an "IV salt" is? (IntraVenous? :laughing: )

README.md
```diff
@@ -678,22 +678,46 @@ The encrypted binary will have the following structure:
...
+            Use ~180MHz ROSC configuration for embedded bootloader
+        --use-mbedtls
+            Use MbedTLS implementation of embedded bootloader
+        --otp-key-page
+            Specify the OTP page storing the AES key (IV salt is stored on the next page)
+        <page>
+            OTP page (default 30)
+        <aes_key>
+            AES Key Share or AES Key
+        <iv_otp>
```
lurch: Would `<iv_salt>` potentially be a less-confusing parameter name than `<iv_otp>`?

README.md
```diff
@@ -678,22 +678,46 @@ The encrypted binary will have the following structure:
...
+            IV OTP Salt
+        <signing_key>
+            Signing Key file
+        <otp>
+            File to save OTP to (will edit existing file if it exists)
```
lurch: Now that the `-t` options have been removed, perhaps this help-text ought to mention what file-format some of these parameters are expected to be in?

README.md
will-v-pi: I've added a better description (it's the Initialisation Vector)

will-v-pi:
Rebased onto the latest develop branch

main.cpp
```diff
@@ -2717,6 +2717,28 @@ uint32_t guess_flash_size(memory_access &access) {
     return size * 2;
 }
 
+// returns true if filename is a hex string, and fills array with the values
+bool filename_to_hex_array(uint8_t idx, uint8_t *array, size_t size) {
+    auto filename = settings.filenames[idx];
```
lurch: IMHO this function would feel more self-contained if it was directly passed the string, instead of being passed an index into an array? :confused:

main.cpp
```diff
@@ -2717,6 +2717,28 @@ uint32_t guess_flash_size(memory_access &access) {
...
+
+    if (!filename.empty() && filename.find("0x") == 0) {
+        // Hex string instead of file
+        if (filename.size() != size*2 + 2) {
+            fail(ERROR_ARGS, "Hex string must be %d characters long (the supplied string is %d characters)", size*2, filename.size() - 2);
```
lurch: Hmmm, maybe this function should also be passed some kind of identifier to be used in the error-messages? :thinking:  Because I guess if you passed both the AES key and also the salt by hex-value, then with this current message you wouldn't know which one of them was wrong!

will-v-pi:
Quick note about TBYB functionality - TLDR is that after being bought, the encrypted binary will need to call `rom_explicit_buy(NULL, 0)` on every boot to cancel the watchdog, but other than that everything works as expected - I think this should be fine

For TBYB to work, it must be set on both the outer decrypting bootloader and the inner encrypted binary. This is because if it is only set on the outer bootloader, then the `chain_image` call with clear the TBYB pending flag, so the inner binary won't call `explicit_buy` correctly.

With it set on both the outer bootloader and the inner binary, on first boot both the bootloader and inner binary will boot with TBYB, and then the inner binary can call `explicit_buy` as normal to clear the flag on the outer bootloader (and write OTP, do erase, etc). It is not possible to clear the flag on the inner binary, as it is encrypted, so you can't just modify one bit.

On subsequent boots, the outer bootloader will boot without TBYB, but the inner binary will have TBYB set by the `chain_image` call - therefore the inner binary will need to call `explicit_buy` every boot, to disable the watchdog. Because this `explicit_buy` will never do anything more than disable the watchdog, you can pass `(NULL, 0)` to it instead of a real buffer, as it does not need a buffer if no flash operations are being performed.

The following would be an example of the TBYB process for an encrypted binary:
```c
boot_info_t boot_info = {};
    int ret = rom_get_boot_info(&boot_info);
    if (ret) {
        if (boot_info.tbyb_and_update_info & BOOT_TBYB_AND_UPDATE_FLAG_BUY_PENDING) {
            uint32_t flash_update_base = boot_info.reboot_params[0];
            if (flash_update_base) {
                // Do self-check etc
            }
            uint32_t buf_size = flash_update_base ? 4096 : 0;
            uint8_t* buffer = flash_update_base ? malloc(buf_size) : NULL;
            int ret = rom_explicit_buy(buffer, buf_size);
            assert(ret == 0);
            if (buffer) free(buffer);
        }
    }
```
as opposed to a normal binary which wouldn't need to check the `flash_update_base` because `BOOT_TBYB_AND_UPDATE_FLAG_BUY_PENDING` will only be set on first boot

will-v-pi:
Rebased onto develop again to pick up mbedtls and workflow fixes

will-v-pi:
Rolled into this I've added permission handling to OTP dump (prints XXs for any rows with permissions failures), and also the option to index by page & row with `--pages` option, rather than the default hex indexing. This makes it much easier to check what pages you've used for keys already.

eg
```
28:40: XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, 
28:48: XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, 
28:56: XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, XXXXXXXX, 
29:00: 0035b631, 000a18d8, 002f2e23, 00027c7b, 0039b1a3, 002890b7, 00072f7b, 0035d241, 
29:08: 0017b551, 002f6203, 001321d6, 0037b50c, 0030178d, 003fd5e6, 00330d6b, 000a8d87, 
29:16: 0011742b, 002bbaa4, 001614b9, 00298875, 0026059b, 0007322d, 002dc151, 003a0935,
```

will-v-pi:
See the [encrypted-shares-ci](https://github.com/raspberrypi/picotool/tree/encrypted-shares-ci) branch for passing CI, which now requires different SDK and examples branches due to the examples build test

bintool/bintool.cpp
```diff
@@ -421,10 +444,32 @@ block place_new_block(std::vector<uint8_t> &bin, uint32_t storage_addr, std::uni
 
     // loop back to first block
     block new_block(new_block_addr, loop_start_rel);
-    // copt the existing block
-    std::copy(first_block->items.begin(),
-              first_block->items.end(),
-              std::back_inserter(new_block.items));
+    // check if last block has an image_def
```
lurch, line 447: From a quick glance, there seems to be a lot of overlap in the change on lines 317-342 and the change on lines 447-472. Does that suggest that a slight refactoring might be appropriate?

bintool/bintool.cpp
```diff
@@ -961,7 +1019,7 @@ int encrypt(elf_file *elf, block *new_block, const private_t aes_key, const publ
 }
 
 
-std::vector<uint8_t> encrypt(std::vector<uint8_t> bin, uint32_t storage_addr, uint32_t runtime_addr, block *new_block, const private_t aes_key, const public_t public_key, const private_t private_key, bool hash_value, bool sign) {
+std::vector<uint8_t> encrypt(std::vector<uint8_t> bin, uint32_t storage_addr, uint32_t runtime_addr, block *new_block, const aes_key_t aes_key, const public_t public_key, const private_t private_key, std::vector<uint8_t> iv_salt, bool hash_value, bool sign) {
```
lurch, line 1022: Why does this `encrypt` function return a `std::vector<uint8_t>`, whilst the `encrypt` function with a similar-looking signature on line 928 just returns an `int` ?

enc_bootloader/CMakeLists.txt
```diff
@@ -0,0 +1,104 @@
...
+        PICO_NO_PROGRAM_INFO=1
+        # No spinlocks used
+        PICO_USE_SW_SPIN_LOCKS=0
+        # No heap is used
+        PICO_HEAP_SIZE=0
+        # These inits are not required
+        PICO_RUNTIME_SKIP_INIT_SPIN_LOCKS_RESET=1
+        PICO_RUNTIME_SKIP_INIT_PER_CORE_IRQ_PRIORITIES=1
+        PICO_BOOTROM_LOCKING_ENABLED=0
+        # PICO_RUNTIME_SKIP_INIT_BOOTROM_LOCKING_ENABLE=1
```
lurch: Why the commented-out setting here? Is it something that ought to be removed, or is it something that the user might want to enable? (and if the latter, maybe a comment would be helpful?)

main.cpp
```diff
@@ -5001,7 +5040,404 @@ vector<uint8_t> sign_guts_bin(iostream_memory_access in, private_t private_key,
...
+    }
+
+    if (get_file_type_idx(1) != get_file_type()) {
+        fail(ERROR_ARGS, "Can only sign to same file type");
+    }
+
+    if (string_to_hex_array(settings.filenames[2], aes_key.bytes, sizeof(aes_key.bytes), "AES key")) {
+        keyFromFile = false;
+    } else if (get_file_type_idx(2) != filetype::bin) {
+        fail(ERROR_ARGS, "Can only read AES key share from BIN file");
```
lurch: I guess technically this ought to be "AES key or AES key share" ?

README.md
```diff
@@ -586,22 +586,46 @@ The encrypted binary will have the following structure:
...
```
lurch: I see that in `main.cpp` this has been updated to "JSON file to save OTP to (will edit existing file if it exists)", so I guess you might want to use #220 here? :grinning:

bintool/bintool.cpp
will-v-pi: There are quite a few functions in bintool which are duplicated for ELFs and BINs, which share some code but with key differences. Some of it could probably be refactored to remove the duplication, but I think that's a change for a different PR, to do that refactoring for all the functions in bintool rather than just the ones changed in this PR.

README.md
```diff
@@ -586,22 +586,46 @@ The encrypted binary will have the following structure:
 - Padding to ensure the encrypted length is a multiple of 4 words
 - Signature metadata block
 
-The AES key must be provided as a .bin file of the 256 bit AES key to be used for encryption.
+The AES key can either be provided as a 1024-bit (128 byte) 4-way key share or as a 256-bit (32 byte) AES key. In the latter case, a key share will be generated for you.
+
+The 4-way key share should be generated by creating a random .bin file of length 128 bytes and the AES key will be derived from that. With the words stored in the file as A[0], B[0], C[0], D[0], A[1], B[1], etc., C[7], D[7], word i of the key X is X[i] = A[i] ^ B[i] ^ C[i] ^ D[i].
+
+Alternatively an AES key can be provided directly, as either a 32 byte .bin file or as a string of 64 hexadecimal characters (eg 0x0123456789abcdeffedcba98765432100123456789abcdeffedcba9876543210). Picotool will generate a random key share for that key, by generating 3 random shares (A, B & C) and calculating a 4th share (D) to match the provided key (X), with each word i of the 4th share calculated as D[i] = X[i] ^ A[i] ^ B[i] ^ C[i].
```
lurch: Maybe it's worth clarifying here that `picotool` will use the size of the .bin file to determine whether the user has supplied an AES key or an AES key-share? And maybe it's also worth stating that it's always the 4-way key share that actually gets written to OTP?

bintool/bintool.cpp
will-v-pi: The BIN versions of most functions returns a vector with the encrypted/signed/etc data, but the ELF versions modify the ELF file in place so just return ints.

README.md
will-v-pi: Yep, I think I'll merge #220 into develop, then merge develop into here again and run it

enc_bootloader/CMakeLists.txt
will-v-pi: It's not required because of the line above - I'll remove it

lurch:
> See the [encrypted-shares-ci](https://github.com/raspberrypi/picotool/tree/encrypted-shares-ci) branch for passing CI, which now requires different SDK and examples branches due to the examples build test

Perhaps the original comment in this PR ought to be updated to say which other pico-sdk and pico-examples PRs ought to be merged before (or in parallel with?) this PR.

bintool/bintool.cpp
will-v-pi: #231

README.md
will-v-pi: I've added those clarifications, although the phrasing might need tidying up

bintool/bintool.cpp
lurch: Aha, thanks :slightly_smiling_face:

README.md
lurch: LGTM!

README.md
```diff
@@ -654,8 +648,8 @@ ENCRYPT:
     Encrypt the program.
 
 SYNOPSIS:
-    picotool encrypt [--quiet] [--verbose] [--embed] [--fast-rosc] [--use-mbedtls] [--otp-key-page <page>] [--hash] [--sign] <infile>
-                [-t <type>] [-o <offset>] <outfile> [-t <type>] <aes_key> <iv_otp> <signing_key> <otp>
+    picotool encrypt [--quiet] [--verbose] [--embed] [--fast-rosc] [--use-mbedtls] [--otp-key-page <page>] [--hash] [--sign] <infile> [-t
+                <type>] [-o <offset>] <outfile> [-t <type>] <aes_key> <iv_salt> <signing_key> <otp>
```
lurch, line 652: How hard would it be to modify the line-wrapping to prevent a line-break getting inserted in the middle of e.g. `[-t <type>]` ?

will-v-pi: I think that's handled by the CLIPP library, so probably a question for @kilograham whether we want to change that code?

https://github.com/raspberrypi/picotool/blob/c56c005bb6490c341183c7620dd45676d26cb602/clipp/clipp.h#L317-L363

lurch: Okay, sounds like There Be Dragons :dragon:

kilograham (changes requested):
Not convinced that's much more helpful. Really thinking about something that helps the user know what happened and what to do - in the RP2040 case it doesn't really make sense to be signing anyway, no?

will-v-pi:
> Not convinced that's much more helpful. Really thinking about something that helps the user know what happened and what to do - in the RP2040 case it doesn't really make sense to be signing anyway, no?

It is a valid use case to be sealing an RP2040 binary, as an RP2350 may want to hash/signature check an RP2040 binary before loading it onto an RP2040. Would this error message be clearer:
```
No metadata block found when sealing RP2040 binary - either use RP2350, or set PICO_CRT0_INCLUDE_PICOBIN_BLOCK=1
```

kilograham (approved):

PR #212, opened 2025-03-09
Reference RP2350-E15 in README

chromy:
Note the specific errata relevant  for the otp permissions command in the README.

---

I assume the 'XXX' errata here is RP2350-E15 ("The bootrom otp_access() function applies incorrect access permission to pages 62 & 63"). I didn't understand the existing comment so I updated it to what I think is meant based on reading the docs but it bears checking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1031,7 +1031,7 @@ Device Descriptor:
 
 ### permissions
 
-This command will run a binary on your device in order to set the OTP permissions, as these are not directly accessible from `picotool` on due to errata RP2350-E15 on RP2350 A2. 
+This command will run a binary on your device in order to set the OTP permissions, as these are not directly accessible from `picotool` due to errata RP2350-E15 on RP2350 A2. 
 Because it runs a binary, the binary needs to be signed if secure boot is enabled. The binary will light an LED when running, which
 can be configured using the `--led` argument. You can define your OTP permissions in a json file, an example of which
 is in [sample-permissions.json](json/sample-permissions.json). The schema for this JSON file is [here](json/schemas/permissions-schema.json)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1031,8 +1031,8 @@ Device Descriptor:
 
 ### permissions
 
-This command will run a binary on your device in order to set the OTP permissions, as these are not directly accessible from `picotool` on due to the default permissions settings required to fix  errata XXX on RP2350. 
-Because it runs a binary, the binary needs to be sign it if secure boot is enabled. The binary will print what it is doing over uart, which
+This command will run a binary on your device in order to set the OTP permissions, as these are not directly accessible from `picotool` on due to errata RP2350-E15 on RP2350 A2.
```
lurch: "on due" -> "due" ?

README.md
```diff
@@ -1031,8 +1031,8 @@ Device Descriptor:
 
 ### permissions
 
-This command will run a binary on your device in order to set the OTP permissions, as these are not directly accessible from `picotool` on due to the default permissions settings required to fix  errata XXX on RP2350. 
-Because it runs a binary, the binary needs to be sign it if secure boot is enabled. The binary will print what it is doing over uart, which
+This command will run a binary on your device in order to set the OTP permissions, as these are not directly accessible from `picotool` on due to errata RP2350-E15 on RP2350 A2. 
+Because it runs a binary, the binary needs to be signed if secure boot is enabled. The binary will print what it is doing over UART, which
 can be configured using the UART Configuration arguments. You can define your OTP permissions in a json file, an example of which
```
will-v-pi: The bit about UART is actually wrong now, as the binary lights an LED instead, so maybe something like the following?

```suggestion
Because it runs a binary, the binary needs to be signed if secure boot is enabled. The binary will light an LED when running, which
can be configured using the `--led` argument. You can define your OTP permissions in a json file, an example of which
```

chromy: Thanks!

README.md
chromy: Thanks! Fixed.

will-v-pi:
Thanks - could you target this PR at the `develop` branch rather than `master`, and then we can merge it

will-v-pi (approved):

PR #213, opened 2025-03-13
Remove holes in ELF files

will-v-pi:
These holes cause issues when signing/hashing/encrypting, as the data in the hole can be different depending on the load method used (00s or FFs)

This will only affect ELF files which are written out afterwards, so the `seal`, `encrypt`, and `partition create` commands.

This replaces and reverts #150 as it's not required and just adds complexity, and fixes https://github.com/raspberrypi/pico-sdk/issues/2321 and https://github.com/raspberrypi/pico-examples/issues/558

### DIFF
--- a/elf/elf_file.cpp
+++ b/elf/elf_file.cpp
@@ -230,15 +230,6 @@ void elf_file::flatten(void) {
         }
         idx++;
     }
-
-    idx = 0;
-    for (const auto &ph : ph_entries) {
-        if (ph.filez) {
-            elf_bytes.resize(std::max(ph.offset + ph.filez, (uint32_t)elf_bytes.size()));
-            memcpy(&elf_bytes[ph.offset], &ph_data[idx][0], ph.filez);
-        }
-        idx++;
-    }
     if (verbose) printf("Elf file size %zu\n", elf_bytes.size());
 }
 
@@ -294,18 +285,6 @@ void elf_file::read_sh_data(void) {
             read_bytes(sh.offset, sh.size, &sh_data[sh_idx][0]);
         }
         sh_idx++;
-    }
-}
-
-void elf_file::read_ph_data(void) {
-    int ph_idx = 0;
-    ph_data.resize(eh.ph_num);
-    for (const auto &ph: ph_entries) {
-        if (ph.filez) {
-            ph_data[ph_idx].resize(ph.filez);
-            read_bytes(ph.offset, ph.filez, &ph_data[ph_idx][0]);
-        }
-        ph_idx++;
     }
 }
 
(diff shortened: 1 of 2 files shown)


elf/elf_file.cpp
```diff
@@ -260,6 +260,29 @@ void elf_file::read_sh(void) {
...
+// If there are holes between sections within segments, increase the section size to plug the hole
+// This is necessary to ensure the whole segment contains data defined in sections, otherwise you end up
+// signing/hashing/encrypting data that may not be written, as many tools write in sections not segments
+void elf_file::remove_sh_holes(void) {
+    for (int i=0; i+1 < sh_entries.size(); i++) {
+        auto sh0 = &(sh_entries[i]);
+        elf32_sh_entry sh1 = sh_entries[i+1];
+        if (
+            (sh0->type == SHT_PROGBITS && sh1.type == SHT_PROGBITS)
+            && (sh0->size && sh1.size)
```
lurch, line 263: Is there a good reason for having `sh0` be a pointer, but `sh1` not be a pointer? IMHO `sh0->size && sh1.size` looks quite weird :wink:

will-v-pi: I did it because `sh0` can be modified but `sh1` doesn't need to be, so `sh0` has to be a pointer but `sh1` can just be a copy

lurch: Fair enough.

elf/elf_file.cpp
```diff
@@ -260,6 +251,29 @@ void elf_file::read_sh(void) {
...
+            && (sh0->addr + sh0->size < sh1.addr)
```
lurch, line 264: Hmmm, I guess you _could_ calculate `uint32_t sh0_endaddr = sh0->addr + sh0->size;` and that'd allow you to replace a few references; but for such simple code it's probably not worth it.

kilograham:
one question; are all the sections always listed in order?

will-v-pi:
> one question; are all the sections always listed in order?

From what I've seen, all sections within the same segment are always in order

kilograham:
ok, next question; are these alignment holes, or bigger holes; e.g. caused by sparse binary. we probably don't want to fill in such gaps there, but hopefully those are sergment gaps? edit: particularly i'm referring to the sorts of weird binaries with holes we've seen in the past

lurch:
> caused by sparse binary. we probably don't want to fill in such gaps there

Would a "sparse binary" also cause similar problems with signing/hashing/encrypting ?

will-v-pi:
Any gaps between sections within a segment (which this PR targets) will always contain corresponding data in the ELF file to fill that gap.

I believe sparse binaries should contain gaps between segments, which are fine because picotool generates a separate load_map entry for each segment. If there are sparse binaries with gaps between sections then these were previously unsupported, and will now work but with the gaps filled with 0s.

kilograham (approved):

PR #219, opened 2025-04-01
Fixup to work with mbedtls 3.x

will-v-pi:
CMake 4.0 has been released, which requires Mbedtls 3.x. This is because Mbedtls 2.x sets a min CMake version of 2.8.12, but CMake 4.0 throws an error if the min version is < 3.5. This is already causing some picotool CI failures, as some runners seem to have the latest CMake (https://github.com/raspberrypi/picotool/actions/runs/14196997735/job/39774407801#step:8:32). This needs updating in the SDK too, see https://github.com/raspberrypi/pico-sdk/issues/2391

The fix is to allow access to the private variables required, and some changed function signatures. This maintains compatibility with mbedtls 2.x using #ifs on the MBEDTLS_VERSION_MAJOR.

### DIFF
--- a/bintool/mbedtls_wrapper.c
+++ b/bintool/mbedtls_wrapper.c
@@ -175,7 +175,11 @@ void mb_sign_sha256(const uint8_t *entropy, size_t entropy_size, const message_d
 
     DEBUG_LOG(" ok (key size: %d bits)\n", (int) ctx_sign.grp.pbits);
 
+#if MBEDTLS_VERSION_MAJOR == 2
     ret = mbedtls_ecp_check_pub_priv(&ctx_sign, &ctx_sign);
+#else
+    ret = mbedtls_ecp_check_pub_priv(&ctx_sign, &ctx_sign, mbedtls_ctr_drbg_random, &ctr_drbg);
+#endif
     DEBUG_LOG("Pub Priv Returned %d\n", ret);
 
     dump_pubkey("  + Public key: ", &ctx_sign);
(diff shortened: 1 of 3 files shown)


lurch:
> This maintains compatibility with mbedtls 2.x using #ifs on the MBEDTLS_VERSION_MAJOR.

If mbedtls 2.x is EOL, and mbed 3.x is also LTS, do we need to bother maintaining compatibility with mbedtls 2.x? :thinking:

will-v-pi:
I think it's best to maintain it for now, so picotool 2.1.x can be built with any SDK 2.1.x

bintool/mbedtls_wrapper.c
```diff
@@ -187,7 +191,11 @@ void mb_sign_sha256(const uint8_t *entropy, size_t entropy_size, const message_d
 
     if ((ret = mbedtls_ecdsa_write_signature(&ctx_sign, MBEDTLS_MD_SHA256,
                                              m->bytes, sizeof(m->bytes),
-                                             out->der, &out->der_len,
+                                             out->der,
+                                             #if MBEDTLS_VERSION_MAJOR >= 3
```
lurch, line 195: Only a minor nitpick, but would it be "safer" to stick to just one style of `#if`, rather than mixing `#if MBEDTLS_VERSION_MAJOR == 2` and `#if MBEDTLS_VERSION_MAJOR >= 3` ?

lurch:
> I think it's best to maintain it for now, so picotool 2.1.x can be built with any SDK 2.1.x

Ahhh, perhaps I was wrong to assume that the pico-sdk and pictool versions increment in lock-step? In which case yeah there's obviously no harm in maintaining compatibility with multiple versions.

bintool/mbedtls_wrapper.c
will-v-pi: I've changed them all to be >= 3

will-v-pi:
> Ahhh, perhaps I was wrong to assume that the pico-sdk and pictool versions increment in lock-step? In which case yeah there's obviously no harm in maintaining compatibility with multiple versions.

They do increment in lock step, but don't necessarily require the latest version of each other - for example picotool 2.1.2 can be built with pico-sdk 2.1.0, if you want the speedups in picotool 2.1.2 but don't want to upgrade your SDK yet. So probably best to maintain compatibility for now.

lurch:
> but don't necessarily require the latest version of each other - for example picotool 2.1.2 can be built with pico-sdk 2.1.0

Ah cool :+1:  With all the intertwined PRs in `picotool`, `pico-sdk` and `pico-examples` recently regarding encryption :closed_lock_with_key: , salts :salt: , etc. I'd falsely assumed that picotool 2.1.2 could only be built with pico-sdk 2.1.2

kilograham (approved):

PR #220, opened 2025-04-01
Add script to update README with output from help commands

will-v-pi:
Add a script that updates these, and an action to check them, which will throw an error if they're not up to date and provide an artifact with the correct README.md. You can then download that artifact and copy it in to replace the README so the action will pass.

Also adds `GENERATE_FIXED_DOCS_WIDTH` CMake option to lock the terminal width to 140, which will generate docs to match the current widths in the README.

This should probably be merged after everything else for 2.1.2, to avoid conflicts in the README file


.github/workflows/check_help_text.yml
```diff
@@ -0,0 +1,55 @@
...
+  check-help-text:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install dependencies (Linux)
+        if: runner.os == 'Linux'
+        run: sudo apt install cmake ninja-build python3 build-essential gcc-arm-none-eabi libnewlib-arm-none-eabi libstdc++-arm-none-eabi-newlib libusb-1.0-0-dev
```
lurch: Do you really need all the cross-build tools just to compile `picotool`?

lurch:
I've not tested this yet, but it sounds like a fantastic idea :heart_eyes:

gen_help_txt.sh
```diff
@@ -0,0 +1,36 @@
...
+
+mkdir -p tmp
+cp README.md tmp/README.md
+
+# Process each help text section
+while IFS= read -r line; do
+    if [[ $line =~ ^\$[[:space:]]*picotool[[:space:]]+help ]]; then
+        # Extract the commands
+        cmd=$(echo "$line" | sed 's/^\$[[:space:]]*//')
+        if [ ! -z "$cmd" ]; then
```
lurch, line 13: This seems to be looking for e.g. `$ picotool help config` in README.md to know which help-text sections to update. But I guess if you add a new command to `picotool` but then forget to add the help-text for that command to README.md, then this script wouldn't flag up any errors?
So what might be super-neat would be for this script to also run `picotool help` and display an error if there are any commands in the help output that don't have their own `picotool help <command>` section in README.md :slightly_smiling_face:  (I guess you'd need to use Bash associative arrays)

.github/workflows/check_help_text.yml
will-v-pi: Nope - I'd just copied it from the other workflow, which does need them as it sets USE_PRECOMPILED=false for some of the tests - will remove them now

gen_help_txt.sh
will-v-pi: I've added that - I did it separately, so the script first updates any existing help text, then parses main.cpp (instead of picotool help, as that had some multiline descriptions that caused problems) to find all commands and checks if they're missing, and if they are it adds their help text to the end for copying/pasting and fails the job. Also added sections for the commands currently missing from the readme.

lurch: Wonderful!

README.md
```diff
@@ -511,6 +514,55 @@ name:      lcd_1602_i2c
 web site:  https://github.com/raspberrypi/pico-examples/tree/HEAD/i2c/lcd_1602_i2c
 ```
 
+## verify
+
+This command allows you to verify the contents of Flash/SRAM on the device matches a file, similar to the `--verify` option for `save` and `load`.
```
lurch: "allows you to verify the contents" -> "allows you to verify that the contents" ?
(and yes, I realise that fixing this will mean editing `main.cpp` rather than directly editing `README.md`)

lurch: Presumably it verifies the contents of Flash by default, and to get it to verify the contents of SRAM you'd need to pass it custom `<from>` and `<to>` addresses?

README.md
```diff
@@ -511,6 +514,55 @@ name:      lcd_1602_i2c
...
+
+```text
+$ picotool help verify
+VERIFY:
+    Check that the device contents match those in the file.
+
+SYNOPSIS:
+    picotool verify [device-selection]
```
lurch: Presumably the `<filename>` argument is mandatory too - should that be displayed here?

README.md
```diff
@@ -511,6 +514,55 @@ name:      lcd_1602_i2c
...
+        -r, --range
+            Compare a sub range of memory only
+        <from>
+            The lower address bound in hex
+        <to>
+            The upper address bound in hex
+        -o, --offset
+            Specify the load address when comparing with a BIN file
+        <offset>
+            Load offset (memory address; default 0x10000000)
```
lurch: Just a suggestion, but I wonder if this would be slightly clearer as:
```
        -o, --offset <offset>
            Specify the load address when comparing with a BIN file
            <offset>
                Load offset (memory address; default 0x10000000)
```
with `--range <from> <to>` being updated similarly?

README.md
```diff
@@ -738,10 +790,29 @@ PARTITION INFO:
     Print the device's partition table.
 
 SYNOPSIS:
-    picotool partition info -m <family_id> [device-selection]
+    picotool partition info [-m <family_id>] [device-selection]
 
 OPTIONS:
-        -m <family_id> [device-selection]
+        -m <family_id>
+            family ID (will show target partition for said family)
```
lurch, line 705: "partition" -> "partition(s)", as I guess there might be multiple partitions with the same family-ID? :shrug:

README.md
```diff
@@ -511,6 +514,55 @@ name:      lcd_1602_i2c
...
+            Filter by product id
+        --ser <ser>
+            Filter by serial number
+        -f, --force
+            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
+            command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
+        -F, --force-no-reboot
+            Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
+            command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but without the
+            RPI-RP2 drive mounted
```
lurch: Does this wording need a slight tweak, as IIRC for RP2350 the drive isn't named `RPI-RP2` ?

README.md
```diff
@@ -933,16 +1003,176 @@ SUB COMMANDS:
...
+
+SYNOPSIS:
+    picotool otp get [-c <copies>] [-r] [-e] [-n] [-i <filename>] [device-selection] [-z] [<selector>..]
+
+OPTIONS:
+    Row/field options
+        -c <copies>
+            Read multiple redundant values
+        -r, --raw
+            Get raw 24 bit values
```
lurch: Small nit, but in general I think "24-bit values" looks better than "24 bit values".

README.md
```diff
@@ -933,16 +1003,176 @@ SUB COMMANDS:
...
+        -z, --fuzzy
+            Allow fuzzy name searches in selector vs exact match
+        <selector>
+            The row/field selector, each of which can select a whole row:
+
+            ROW_NAME to select a whole row by name.
+            ROW_NUMBER to select a whole row by number.
+            PAGE:PAGE_ROW_NUMBER to select a whole row by page and number within page.
+
+            ... or can select a single field/subset of a row (where REG_SEL is one of the above row selectors):
```
lurch: I know very little about OTP, but would `ROW_SEL` be better than `REG_SEL`?

README.md
```diff
@@ -933,16 +1003,176 @@ SUB COMMANDS:
...
+
+            REG_SEL.FIELD_NAME to select a field within a row by name.
+            REG_SEL.n-m to select a range of bits within a row.
+            REG_SEL.n to select a single bit within a row.
+            .FIELD_NAME to select any row's field by name.
+
+            .. or can selected multiple rows by using blank or '*' for PAGE or PAGE_ROW_NUMBER
```
lurch: selected -> select ?

README.md
```diff
@@ -933,16 +1003,176 @@ SUB COMMANDS:
...
+            ROW_NAME to select a whole row by name.
+            ROW_NUMBER to select a whole row by number.
+            PAGE:PAGE_ROW_NUMBER to select a whole row by page and number within page.
+
+            ... or can select a single field/subset of a row (where REG_SEL is one of the above row selectors):
+
+            REG_SEL.FIELD_NAME to select a field within a row by name.
+            REG_SEL.n-m to select a range of bits within a row.
+            REG_SEL.n to select a single bit within a row.
+            .FIELD_NAME to select any row's field by name.
```
lurch, line 956: Does this mean it'd print all rows that contain a matching FIELD_NAME ?

README.md
```diff
@@ -933,16 +1003,176 @@ SUB COMMANDS:
...
+        -s, --set-bits
+            Set bits only
+        -i <filename>
+            Include extra otp definition
+        <value>
+            The value to set
+    Row/Field Selection
+        -z, --fuzzy
+            Allow fuzzy name searches in selector vs exact match
+        <selector>
```
lurch, line 1006: Is the `selector` functionality here the same as for `otp get`? In which case perhaps the corresponding help-text should be the same too? (I'll let you decide which of the two variants is better)
(Or if not the same, at least more similar to each other?)

README.md
```diff
@@ -933,16 +1003,176 @@ SUB COMMANDS:
...
+```
+
 ### load
 
 This command allows loading of a range of OTP rows onto the device. The source can be a binary file, or a JSON file such as the one output by `picotool sign`. The schema for this JSON file is [here](json/schemas/otp-schema.json)
+
+```text
+$ picotool help otp load
+OTP LOAD:
+    Load the row range stored in a file into OTP and verify. Data is 2 bytes/row for ECC, 4 bytes/row for raw.
```
lurch: > 4 bytes/row for raw

Will the 4th byte simply get ignored? Does it ignore the MSB or LSB?

README.md
```diff
@@ -933,16 +1003,176 @@ SUB COMMANDS:
...
+
+SYNOPSIS:
+    picotool otp load [-r] [-e] [-s <row>] [-i <filename>] <filename> [-t <type>] [device-selection]
+
+OPTIONS:
+    Row options
+        -r, --raw
+            Get raw 24 bit values
```
lurch: Get -> Set ?

README.md
```diff
@@ -933,16 +1003,176 @@ SUB COMMANDS:
...
+        -e, --ecc
+            Use error correction
```
lurch: As already mentioned, I'm not very familiar with OTP, but are the `-r` and `-e` options mutually exclusive? If neither is specified, does it default to raw or ecc mode? (Or am I just totally misunderstanding?!?)

README.md
will-v-pi: No, it will just give you the target partition for that family ID, as if you'd dragged and dropped a UF2 with that ID. It'll do all the A/B stuff etc, and just return one partition

README.md
```diff
@@ -933,16 +1003,176 @@ SUB COMMANDS:
...
+        -s <row>
+            Start row to load at (note use 0x for hex)
```
lurch, line 1060: So (unlike `otp set`) there's no option to identify rows by name?

README.md
```diff
@@ -1102,6 +1331,149 @@ Loading into XIP RAM: [==============================]  100%
...
+            PAGE:PAGE_ROW_NUMBER to select a whole row by page and number within page.
+
+            ... or can select a single field/subset of a row (where REG_SEL is one of the above row selectors):
+
+            REG_SEL.FIELD_NAME to select a field within a row by name.
+            REG_SEL.n-m to select a range of bits within a row.
+            REG_SEL.n to select a single bit within a row.
+            .FIELD_NAME to select any row's field by name.
+
+            .. or can selected multiple rows by using blank or '*' for PAGE or PAGE_ROW_NUMBER
```
lurch: Again, 'selected' -> 'select'

README.md
```diff
@@ -1102,6 +1331,149 @@ Loading into XIP RAM: [==============================]  100%
...
+
+This command is used by the SDK to post-process dissassembly files, to replace the coprocessor instructions (mcr, mrc, etc.) with more human readable ones (eg rcp_bvalid, gpioc_hi_put, etc.).
+
+```text
+$ picotool help coprodis
+COPRODIS:
+    Post-process coprocessor instructions in disassembly files.
+
+SYNOPSIS:
+    picotool coprodis [--quiet] [--verbose] <infile> [-t <type>] <outfile> [-t <type>]
```
lurch, line 1335: Two separate `-t` flags on the same command-line? Does that work okay?

README.md
```diff
@@ -1102,6 +1331,149 @@ Loading into XIP RAM: [==============================]  100%
...
+    Input DIS
+        <infile>
+            The file name
+        -t <type>
+            Specify file type (uf2 | elf | bin) explicitly, ignoring file extension
+    Output DIS
+        <outfile>
+            The file name
+        -t <type>
+            Specify file type (uf2 | elf | bin) explicitly, ignoring file extension
```
lurch, line 1351: A "uf2" is also a disassembly file?

README.md
will-v-pi: Yeah, the CLI stuff is all a bit weird - this is not a change I know how to make.

README.md
```diff
@@ -1102,6 +1331,149 @@ Loading into XIP RAM: [==============================]  100%
...
+```
+
+## link
+
+This command is used to link multiple binaries with block loops into a single larger block loop. It can currently link up to 3 files into a block loop. It will add the required Rolling Window Delta items to the new block loop, to ensure that everyting is rolled correctly when being executed. For examples of it's usage, see the universal examples in pico-examples.
```
lurch: nit: `it's` -> `its`

I guess there'd be no harm in adding a direct link to https://github.com/raspberrypi/pico-examples?tab=readme-ov-file#universal here?

README.md
```diff
@@ -1102,6 +1331,149 @@ Loading into XIP RAM: [==============================]  100%
...
+
+```text
+$ picotool help link
+LINK:
+    Link multiple binaries into one block loop.
+
+SYNOPSIS:
+    picotool link [--quiet] [--verbose] <outfile> [-t <type>] <infile1> [-t <type>] <infile2> [-t <type>] [<infile3>] [-t <type>] [-p] <pad>
```
lurch: I wonder if the `<outfile> [-t <type>] <infile1> [-t <type>] <infile2> [-t <type>] [<infile3>] [-t <type>]` here could/should be `<outfile> [-t <out_type>] <infile1> [-t <in_type>] <infile2> [-t <in_type>] [<infile3> [-t <in_type>]]` which might then make the following help-text easier to understand?

README.md
lurch: In hindsight (having looked at the other help sections), perhaps the "problem" here is that `picotool verify [device-selection]` is simply too compact, and isn't showing all the available options being used, in contrast to e.g. `picotool erase -r <from> <to> [device-selection]` ? :thinking:

README.md
will-v-pi: Will fix this in #207, which adds `named_untyped_file_selection`

will-v-pi: Done 338499c84725e74da5492c959e440afeb692e489

README.md
will-v-pi: It'll verify the contents matches whatever file you pass it - so if you pass it a no_flash ELF or UF2 file then it'll verify against SRAM as those file types encode the data and memory address it goes to, but if you pass it a BIN file you need to specify the `--offset` as BIN files only contain the data

README.md
will-v-pi: It ignores the MSB - I'll add a note to that line

README.md
will-v-pi: `dump` and `load` default to raw unless you specify ecc - for others it will use the default for the row, and you can override it with `-r` or `-e`. I'll add a note to `dump` and `load` about the default being raw

README.md
will-v-pi: No, that's not an option, as this is more for loading raw data into OTP, rather than configuring the device using OTP

README.md
will-v-pi: Yes, it seems to work fine, and associates the passed `-t <type>` with the immediately preceding file name

README.md
```diff
@@ -211,7 +211,7 @@ TARGET SELECTION:
         -F, --force-no-reboot
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but without the
-            RPI-RP2 drive mounted
```
lurch, line 119: I dunno if it's "technically correct" but in some other places I refer to this as the "BOOTSEL drive", which _might_ be less ambiguous than "USB drive"? :shrug:

README.md
lurch: Will `picotool` give you an error if you try using both the `-r` and `-e` options at the same time?

will-v-pi: No, instead `-r` will trump `-e` currently - I think it's fine to leave it that way, as it gives you the option to change the default but still use both (eg add some bash profile to always call `-e`, but you can still call `-r` as well to override it)

lurch: Okay, and does it _tell_ you that it's doing `raw` instead of `ecc`? Or does the user have to guess? (apologies for the awkward questions, I'm just trying to prevent future user-confusion)

will-v-pi: It doesn't tell you explicity, but `dump` will show 16-bit values for ecc and 32-bit values for raw, and `load` will throw an error if your file isn't aligned to what you select, or if it's too big cause you're passing an ecc file to a raw load.

The load command with a .bin file is quite a dangerous one anyway, so any users using it should really understand exactly what it is doing (and realise they can't do both ecc and raw), else they can easily brick their device.

lurch: Okay - I'll stop with the picky questions and assume that people using OTP know what they're doing then :grinning:

kilograham (approved):
Note (mostly unrelated); i'm not sure if this is captured elsewhere, but in some cases I think the `-t type` is superfluous (there are some things which only have one type i think)

will-v-pi:
> Note (mostly unrelated); i'm not sure if this is captured elsewhere, but in some cases I think the `-t type` is superfluous (there are some things which only have one type i think)

Yep, that's fixed in #207

README.md
```diff
@@ -837,20 +907,184 @@ SUB COMMANDS:
...
+OTP SET:
+    Set the value of an OTP row/field
+
+SYNOPSIS:
+    picotool otp set [-c <copies>] [-r] [-e] [-s] [-i <filename>] [-z] <selector> <value> [device-selection]
+
+OPTIONS:
+    Redundancy/Error Correction Overrides
+        -c <copies>
+            Read multiple redundant values
```
lurch: Is it correct for this to say "**Read** multiple redundant values" given that this is the "otp **set**" command?

README.md
```diff
@@ -1010,6 +1243,149 @@ Loading into XIP RAM: [==============================]  100%
...
+
+This command is used to link multiple binaries with block loops into a single larger block loop. It can currently link up to 3 files into a block loop. It will add the required Rolling Window Delta items to the new block loop, to ensure that everyting is rolled correctly when being executed. For examples of its usage, see the universal examples in [pico-examples](https://github.com/raspberrypi/pico-examples?tab=readme-ov-file#universal).
+
+```text
+$ picotool help link
+LINK:
+    Link multiple binaries into one block loop.
+
+SYNOPSIS:
+    picotool link [--quiet] [--verbose] <outfile> [-t <type>] <infile1> [-t <type>] <infile2> [-t <type>] [<infile3>] [-t <type>] [-p] <pad>
```
lurch: Is the `[-p] <pad>` here supposed to be `[-p <pad>]` ?

PR #221, opened 2025-04-02
git: ignore picotool installation directory

stefanb2:
Pico SDK build option PICOTOOL_FETCH_FROM_GIT_PATH expects picotool build to be executed with

        -DCMAKE_INSTALL_PREFIX=. -DPICOTOOL_FLAT_INSTALL=1

i.e. the files get installed under ./picotool/. Tell git to ignore this directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,8 @@ bazel-*
 
 # Ignore until https://github.com/bazelbuild/bazel/issues/20369 is fixed.
 MODULE.bazel.lock
+
+# Pico SDK expects ${PICOTOOL_FETCH_FROM_GIT_PATH}/picotool and
+#    cmake -DCMAKE_INSTALL_PREFIX=. -DPICOTOOL_FLAT_INSTALL=1 ...
+# i.e. the files get installed under ./picotool/.
+/picotool/*


will-v-pi:
This isn't needed - when the SDK downloads picotool to `PICOTOOL_FETCH_FROM_GIT_PATH`, the actual git repo is in `PICOTOOL_FETCH_FROM_GIT_PATH/picotool-src` and it gets installed to `PICOTOOL_FETCH_FROM_GIT_PATH/picotool`, so the installation is out of the picotool-src tree and therefore the gitignore isn't needed.

See the pico-sdk/tools/Findpicotool.cmake file for how it works - note that you should not set `PICOTOOL_FETCH_FROM_GIT_PATH` to an existing clone of picotool (which might be what's causing this issue here), you should set it to an empty directory and then let the SDK clone and build picotool in that directory.

stefanb2:
Things would be much easier if `Findpicotool.cmake` would simply look into `PATH` first.

lurch:
@stefanb2 See https://github.com/raspberrypi/picotool?tab=readme-ov-file#usage-by-the-raspberry-pi-pico-sdk , if you haven't already.

PR #226, opened 2025-04-11
Use Mbedtls files directly

will-v-pi:
Update to a more pico-sdk style inclusion of mbedtls, by including the files directly

Tested against develop (mbedtls 2.x), and https://github.com/raspberrypi/pico-sdk/pull/2405 (mbedtls 3.x)


lib/BUILD.bazel
```diff
@@ -1,6 +1,6 @@
 cc_library(
     name = "mbedtls_config",
-    hdrs = ["include/mbedtls_config.h"],
+    hdrs = ["include/mbedtls_config.h", "include/picotool_mbedtls_config.h"],
```
lurch, line 3: Does it make sense to list **both** of these files, when `mbedtls_config.h` is merely a wrapper around `picotool_mbedtls_config.h` ? Or is this a required backwards-compatibility-thing in order not to break any existing Bazel projects? (I'm afraid I don't know much about Bazel!)

armandomontanez: If both are going to be referenced, both need to be added. This is because Bazel sandboxes things, so if a file is not listed it won't be available during compilation.

With that said, this has made me realize that the mbedtls configuration handling is wrong in Bazel, so don't worry too much about it. (you're welcome to file an issue and assign it to me)

lib/CMakeLists.txt
```diff
@@ -21,7 +21,121 @@ add_subdirectory(nlohmann_json EXCLUDE_FROM_ALL)
 add_subdirectory(whereami EXCLUDE_FROM_ALL)
 
 if(EXISTS "${PICO_SDK_PATH}/lib/mbedtls/CMakeLists.txt")
-    option(ENABLE_PROGRAMS "Build Mbed TLS programs." OFF)
-    option(ENABLE_TESTING "Build Mbed TLS tests." OFF)
-    add_subdirectory(${PICO_SDK_PATH}/lib/mbedtls mbedtls EXCLUDE_FROM_ALL)
+    set(PICO_MBEDTLS_PATH ${PICO_SDK_PATH}/lib/mbedtls)
```
kilograham, line 31: I think the user should be able to specify/override `PICO_MBEDTLS_PATH` still

will-v-pi: I've added the initialization from the SDK, so you can set it with a CMake or environment variable

lurch:
~~CI is failing with `ERROR: Unknown command: seal` :seal:~~
EDIT: CI now passing :partying_face:

kilograham (approved):

PR #230, opened 2025-04-22
Add instructions to udev rules file, and fix order

will-v-pi:
Adding the instructions into the file is useful for linking to the file from elsewhere without needing to explain where to copy it, eg from the Pico VS Code extension

Order fix supercedes #199

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -30,7 +30,7 @@ make
 On Linux you can add udev rules in order to run picotool without sudo:
 
 ```console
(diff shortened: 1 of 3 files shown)


lurch (approved):

lurch (approved):

kilograham:
should we not delete `99-`?

will-v-pi:
> should we not delete `99-`?

This renames `99-` to `60-`

kilograham:
oops; duh - it didn't pop out as such in the git web interface, but i see it now!

PR #232, opened 2025-04-29
Minor tidyups

will-v-pi:
Don't add a `hash_def` item when not signing/hashing (currently it's added whenever `picotool seal` is invoked)

Validate that the partition name is < 128 characters long when creating partitions, as the size must fit within 7 bits

### DIFF
--- a/bintool/bintool.cpp
+++ b/bintool/bintool.cpp
@@ -546,6 +546,11 @@ uint32_t calc_checksum(std::vector<uint8_t> bin) {
 
 #if HAS_MBEDTLS
 void hash_andor_sign_block(block *new_block, const public_t public_key, const private_t private_key, bool hash_value, bool sign, std::vector<uint8_t> to_hash) {
+    if (!(hash_value || sign)) {
+        // Don't need to add anything if not actually hashing or signing
+        return;
+    }
+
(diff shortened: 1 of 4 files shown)


bintool/bintool.cpp
```diff
@@ -546,6 +546,11 @@ uint32_t calc_checksum(std::vector<uint8_t> bin) {
 
 #if HAS_MBEDTLS
 void hash_andor_sign_block(block *new_block, const public_t public_key, const private_t private_key, bool hash_value, bool sign, std::vector<uint8_t> to_hash) {
+    if (!(hash_value || sign)) {
```
lurch, line 549: If neither `hash_value` nor `sign` are set, then perhaps it'd make more sense to just not call `hash_andor_sign_block` at all, rather than calling it and early-exiting? :shrug:

will-v-pi: That requires duplicate code in the 3 places it's called, so I think this is neater - context is that the `hash_andor_sign` functions for BINs and ELFs now also have a valid use case of adding a load_map without any hashing/signing (eg for a packaged SRAM binary), so the `hash_andor_sign` functions can be used to do both, either, or neither

lurch: "so I think this is neater" - OK :)

kilograham (approved):

PR #237, opened 2025-06-02
Ignore other blocks when signing with rollback version

will-v-pi:
This sets any other blocks (other than partition tables) as ignored blocks when signing with a rollback version, to avoid hitting RP2350-E13

Bug was raised in https://forums.raspberrypi.com/viewtopic.php?t=388370


kilograham (commented):
looks good but i think "ignore_others" is confusing as it is "make_others_ignored"

kilograham (approved):

PR #240, opened 2025-06-04
better hardening of enc_bootloader for fault injection

kilograham:


main.cpp
```diff
@@ -4923,6 +4923,12 @@ void sign_guts_elf(elf_file* elf, private_t private_key, public_t public_key) {
         image_type->flags |= PICOBIN_IMAGE_TYPE_EXE_TBYB_BITS;
     }
 
+    if (settings.seal.set_tbyb) {
```
kilograham: note this was kicking around in the branch - @will-v-pi not sure if it belongs or not

will-v-pi: No, it's just a duplication of the code directly above it, probably due to merging fun

enc_bootloader/enc_bootloader.elf
will-v-pi, line 1: These ELF files contain debug info, including local paths - I think these should be disabled by the `-DPICO_DEBUG_INFO_IN_RELEASE=OFF` flag when creating the ExternalProjects?

will-v-pi:
Tested and everything's running correctly

enc_bootloader/enc_bootloader.elf
kilograham: Yeah, i did it manually, because the build was failing from the top-level build, but i see now that was a stale sub-build directory,

That said, it would be nice if the top-level (non USE_PRE_COMPILE) build copied the binaries over

will-v-pi: Have added that in #242

will-v-pi (approved):

PR #242, opened 2025-06-11
Install pre-compiled files into source tree

will-v-pi:
This adds an install step in the external projects when `USE_PRECOMPILED=false` to copy the compiled binaries into the source tree.

This also adds a new action to check that the pre-compiled files have been updated correctly, although I'm not sure if we should add that or not, as the files will change slightly depending on the compiler/SDK used so that action will probably fail quite a lot?

Fixes https://github.com/raspberrypi/picotool/pull/240#discussion_r2129222486


.github/workflows/check_precompiled.yml
```diff
@@ -0,0 +1,81 @@
...
+jobs:
+  check-precompiled:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install dependencies (Linux)
+        if: runner.os == 'Linux'
```
lurch: Is `if: runner.os == 'Linux'` needed, given that you also have `runs-on: ubuntu-latest` ?

lurch:
Looks like there are merge conflicts (which are probably what's also causing the CI failure?)

will-v-pi:
That CI failure is why I'm thinking this change doesn't want to have the check action, as the only change has been in the SDK so we don't really need to update the `.elf`

The conflict is just because the code for those files has been updated

.github/workflows/check_precompiled.yml
```diff
@@ -23,50 +26,65 @@ jobs:
...
+          # enc_bootloader
+          enc_bootloader_elf_not_updated=$(! echo "$updated_files" | grep -q "enc_bootloader/enc_bootloader.elf")
+          enc_bootloader_mbedtls_elf_not_updated=$(! echo "$updated_files" | grep -q "enc_bootloader/enc_bootloader_mbedtls.elf")
+          echo "enc_bootloader_elf_not_updated: $enc_bootloader_elf_not_updated"
+          echo "enc_bootloader_mbedtls_elf_not_updated: $enc_bootloader_mbedtls_elf_not_updated"
+          for file in enc_bootloader/*; do
+            if echo "$updated_files" | grep -q "$file"; then
+              echo "File $file is in the PR"
+              if $enc_bootloader_elf_not_updated || $enc_bootloader_mbedtls_elf_not_updated; then
+                echo "File $file is in the PR but not enc_bootloader/enc_bootloader.elf"
```
lurch: Should this be "...but not enc_bootloader/enc_bootloader.elf or enc_bootloader/enc_bootloader_mbedtls.elf" ?

.github/workflows/check_precompiled.yml
```diff
@@ -23,50 +26,65 @@ jobs:
...
-            ALL_FINE=0
-          fi
+          echo "Updated files: $updated_files"
 
-          if ! cmp -s picoboot_flash_id/flash_id.bin picoboot_flash_id/flash_id.bin.orig; then
-            echo "Error: flash_id.bin is out of date"
-            ALL_FINE=0
-          fi
+          # enc_bootloader
+          enc_bootloader_elf_not_updated=$(! echo "$updated_files" | grep -q "enc_bootloader/enc_bootloader.elf")
```
lurch: Might be worth tightening up your grep regexes slightly?
```console
$ echo "FOOenc_bootloader/enc_bootloaderXelfBAR" | grep "enc_bootloader/enc_bootloader.elf"
FOOenc_bootloader/enc_bootloaderXelfBAR
```
:wink:

.github/workflows/check_precompiled.yml
```diff
@@ -0,0 +1,97 @@
...
+          echo "enc_bootloader_mbedtls_elf_not_updated: $enc_bootloader_mbedtls_elf_not_updated"
+          for file in enc_bootloader/*; do
+            if echo "$updated_files" | grep -q "$file"; then
+              echo "File $file is in the PR"
+              if $enc_bootloader_elf_not_updated || $enc_bootloader_mbedtls_elf_not_updated; then
+                echo "File $file is in the PR but not enc_bootloader/enc_bootloader.elf"
+                exit 1
+              fi
+            else
+              echo "File $file is not in the PR"
```
lurch: Not sure there's much use in printing out filenames which aren't in the PR?

.github/workflows/check_precompiled.yml
```diff
@@ -0,0 +1,97 @@
...
+          # picoboot_flash_id
+          flash_id_bin_not_updated=$(! echo "$updated_files" | grep -q "picoboot_flash_id/flash_id.bin")
+          echo "flash_id_bin_not_updated: $flash_id_bin_not_updated"
+          for file in picoboot_flash_id/*; do
+            if echo "$updated_files" | grep -q "$file"; then
+              echo "File $file is in the PR"
+              if $flash_id_bin_not_updated; then
+                echo "File $file is in the PR but not picoboot_flash_id/flash_id.bin"
+                exit 1
+              fi
```
lurch: ~~If I'm understanding the code correctly, then it looks like you could skip creating the `$flash_id_bin_not_updated` variable, and just change the "if" line to~~
```
if $file != "picoboot_flash_id/flash_id.bin"
```
?
~~(IMHO more concise code is generally more readable)~~

EDIT: Oh, now I realise that I was misled by your error message :facepalm: I guess you actually meant something like:
```
echo "File $file is modified in the PR but picoboot_flash_id/flash_id.bin hasn't been updated"
```
?

.github/workflows/check_precompiled.yml
will-v-pi: Sorry, this was just a test commit, hence all the printing - I'll let you know when it's working again for review

.github/workflows/check_precompiled.yml
will-v-pi: I've anchored the starts & ends

.github/workflows/check_precompiled.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+# Pass updated files in $updated_files environment variable
+
+# enc_bootloader
+echo "$updated_files" | grep -q "^enc_bootloader/.*\.elf$"
+enc_bootloader_elf_not_updated=$?
```
lurch: Should this check that _both_ of the enc_bootloader ELFs have been updated? As it is, I think this check would pass if only one of the ELFs was updated but the other wasn't?

will-v-pi: That was intentional, because you might only change files that are used by one of the versions - this action is mostly a quick double check that you tried to upload changed ELF files, so if you're uploading one and both have changed, you'll hopefully remember to upload the other.

.github/workflows/check_precompiled.sh
```diff
@@ -0,0 +1,51 @@
...
+if [ $enc_bootloader_elf_not_updated -eq 1 ]; then
+    echo "Checking enc_bootloader files for modifications as ELFs have not been updated"
+    for file in enc_bootloader/*; do
+        if echo "$file" | grep -q "CMakeLists.txt" || echo "$file" | grep -q "BUILD.bazel"; then
```
lurch: I _think_ that this could be simplified to
```shell
if [[ "$file" == "enc_bootloader/CMakeLists.txt" ]] || [[ "$file" == "enc_bootloader/BUILD.bazel" ]]; then
```
?

enc_bootloader/enc_bootloader_mbedtls.elf
lurch, line 1: Is the file-permissions change here intentional?

will-v-pi: Thanks for spotting that - it's because of the install step used, I will change that so it installs them all as files rather than programs. Although I'm not sure why all of them aren't showing up as changed modes, as they were all changed to executable like this?

lurch: > Although I'm not sure why all of them aren't showing up as changed modes, as they were all changed to executable like this?

Looks like some of them were already 755, and they've now **all** been set back to 644 :slightly_smiling_face:

.github/workflows/check_precompiled.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+# Pass updated files in $updated_files environment variable
```
lurch, line 3: _Conventionally_ environment variables are always created as `$UPPERCASE_VARIABLES`, but for such a simple one-off CI script I guess this doesn't really matter.

.github/workflows/check_precompiled.sh
```diff
@@ -0,0 +1,51 @@
...
+# enc_bootloader
+echo "$updated_files" | grep -q "^enc_bootloader/.*\.elf$"
+enc_bootloader_elf_not_updated=$?
+if [ $enc_bootloader_elf_not_updated -eq 1 ]; then
+    echo "Checking enc_bootloader files for modifications as ELFs have not been updated"
+    for file in enc_bootloader/*; do
+        if [[ "$file" == "enc_bootloader/CMakeLists.txt" || "$file" == "enc_bootloader/BUILD.bazel" ]]; then
+            continue
+        fi
+        if echo "$updated_files" | grep -q "$file"; then
```
lurch: If you wanted to avoid the use of all these `grep -q` calls, I guess you could use the [builtin bash functionality](https://stackoverflow.com/a/229606) instead (and there's also a [regex-version](https://stackoverflow.com/a/231298) that you could use on e.g. line 6 above).

lurch:
Hmmm, something that's just occurred to me: if a change was made to a comment in one of the `.c` files, I guess that might result in no change to the generated ELF file; and the CI checks for that comment-fix PR might fail? :thinking:  (but I guess in such a scenario we could just ignore the failing CI and merge anyway?)

will-v-pi:
> Hmmm, something that's just occurred to me: if a change was made to a comment in one of the `.c` files, I guess that might result in no change to the generated ELF file; and the CI checks for that comment-fix PR might fail? 🤔 (but I guess in such a scenario we could just ignore the failing CI and merge anyway?)

Yeah, that's something to bear in mind, but I think that should be fine

.github/workflows/check_precompiled.sh
will-v-pi: I wanted to treat this more as a local variable anyway - it's only an environment variable to avoid any argument passing issues

.github/workflows/check_precompiled.sh
will-v-pi: That builtin regex functionality doesn't seem to work for the multiline `$updated_files` string, which has one filename per line
(eg `[[ "$updated_files" =~ "^enc_bootloader/.*\.elf$" ]]` always returns `false`)

lurch: Hmmm, perhaps the `^` and `$` are matching start-of-string and end-of-string, rather than start-of-line and end-of-line?
Doesn't really matter though, keep using your existing solution :slightly_smiling_face:

.github/workflows/check_precompiled.sh
```diff
@@ -3,47 +3,41 @@
...
-if [ $flash_id_bin_not_updated -eq 1 ]; then
+if ! echo "$updated_files" | grep -q "^picoboot_flash_id/.*\.bin$"; then
     echo "Checking picoboot_flash_id files for modifications as BINs have not been updated"
     for file in picoboot_flash_id/*; do
         if [[ "$file" == "picoboot_flash_id/CMakeLists.txt" || "$file" == "picoboot_flash_id/BUILD.bazel" ]]; then
             continue
         fi
-        if echo "$updated_files" | grep -q "$file"; then
+        if [[ "$updated_files" == *"$file"* ]]; then
             echo "File $file is in the PR but flash_id BINs have not been updated"
```
lurch, line 27: Minor nitpick - is there more than one flash_id BIN?

.github/workflows/check_precompiled.sh
```diff
@@ -3,47 +3,41 @@
...
-if [ $xip_ram_perms_elf_not_updated -eq 1 ]; then
+if ! echo "$updated_files" | grep -q "^xip_ram_perms/.*\.elf$"; then
     echo "Checking xip_ram_perms files for modifications as ELFs have not been updated"
     for file in xip_ram_perms/*; do
         if [[ "$file" == "xip_ram_perms/CMakeLists.txt" || "$file" == "xip_ram_perms/BUILD.bazel" ]]; then
             continue
         fi
-        if echo "$updated_files" | grep -q "$file"; then
+        if [[ "$updated_files" == *"$file"* ]]; then
             echo "File $file is in the PR but xip_ram_perms ELFs have not been updated"
```
lurch, line 41: Minor nitpick - is there more than one xip_ram_perms ELF?

.github/workflows/check_precompiled.sh
will-v-pi: Not now, but there might be in the future

.github/workflows/check_precompiled.sh
will-v-pi: Not now, but there might be in the future

CMakeLists.txt
```diff
@@ -62,6 +62,12 @@ if (NOT DEFINED USE_PRECOMPILED)
     set(USE_PRECOMPILED true)
 endif()
 
+if (USE_PRECOMPILED)
+    set(EXTERNAL_PROJECT_INSTALL_COMMAND "")
+else()
+    set(EXTERNAL_PROJECT_INSTALL_COMMAND ${CMAKE_COMMAND} --install .)
```
kilograham: "," gives me the willies

will-v-pi: I've removed that and added a better fix, with a new CMake function to handle the install vs don't install cases

kilograham (approved):

PR #243, opened 2025-06-13
Fix encryption when using clang, and with allocatable heap

will-v-pi:
This fixes encryption when using clang by plugging gaps between segments when encrypting, in addition to already plugging gaps between sections within segments when signing/hashing/encrypting.

Also fixes encryption with allocatable heap, by moving the signature block into the heap for no_flash binaries (as it already is with GCC), to ensure that the signature comes at the end of a contiguous region of defined data, as encryption currently requires a contiguous region. This will fix https://github.com/raspberrypi/pico-sdk/pull/2515


will-v-pi:
Also added a test build of pico-examples using latest clang, which would have caught this

kilograham (approved):

PR #246, opened 2025-06-20
Make no libusb build clearer

### LINKED ISSUE #245: Unhelpful error message when compiled without USB support
The documentation for picotool mentions that libusb-1.0-0-dev should be installed before compiling picotool. Unfortunately, the manual configuration instructions in the "Getting started with Raspberry Pi Pico-series" do _not_ mention this, and seem to just assume you already have a working install of picotool.

If this step is missed, picotool will build successfully but without USB support. The executable then silently behaves differently to one which was built with USB support, resulting in misleading error messages when trying to use multi-purpose commands like `picotool info`.

When built with USB support, running `picotool info` without any further arguments is allowed, and lists any attached Pico devices.

When built without USB support, `picotool info` fails, and prints an error message about a missing required argument - because it is only expecting to be run on a firmware file:

```
$ picotool info

ERROR: missing <filename>

SYNOPSIS:
    picotool info [-b] [-m] [-p] [-d] [--debug] [-l] [-a] <filename> [-t <type>]

Use "picotool help info" for more info
```

To add to the confusion, the message printed by `picotool help` still says "Without any arguments, this will display basic information for all RP-series devices in BOOTSEL mode", even though in this configuration that isn't true!

Please can you add a specific error message for this case? Eg, if `picotool info` (and the other dual-purpose commands) without arguments printed a message to the effect of "this version of picotool was compiled without USB support, so cannot search for attached devices", that would make the issue much clearer.

will-v-pi:
Fix info and config help text without libusb, add no USB support messages to command line parsing errors, and add no USB support message to `picotool version`

```
$ picotool version
picotool v2.1.2-develop (Linux, GNU-12.2.0, RelWithDebInfo)

This version of picotool was compiled without USB support. Some commands are not available.
```

Eg, the output from `picotool info` is now
```
$ picotool info

ERROR: missing <filename>

SYNOPSIS:
    picotool info [-b] [-m] [-p] [-d] [--debug] [-l] [-a] <filename> [-t <type>]

Use "picotool help info" for more info

This version of picotool was compiled without USB support. Some commands are not available.
```
or `picotool load`
```
$ picotool load

ERROR: Unknown command: load

COMMANDS:
    info        Display information from the target file.
    config      Display or change program configuration settings from the target file.
    encrypt     Encrypt the program.
    seal        Add final metadata to a binary, optionally including a hash and/or signature.
    link        Link multiple binaries into one block loop.
    otp         Commands related to the RP2350 OTP (One-Time-Programmable) Memory
    partition   Commands related to RP2350 Partition Tables
    uf2         Commands related to UF2 creation and status
    version     Display picotool version
    coprodis    Post-process coprocessor instructions in disassembly files.
    help        Show general help or help for a specific command

Use "picotool help <cmd>" for more info

This version of picotool was compiled without USB support. Some commands are not available.
```

Fixes #245

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -8,7 +8,7 @@ git submodule update --init lib/mbedtls
 
 You also need to install `libusb-1.0` if you want to use the USB functionality.
 
-> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands won't appear in the output of the help command. The build output message `libUSB is not found - no USB support will be built` also appears in the build logs, and if you attempt to execute an invalid command you get the error `This version of picotool was compiled without USB support. Some commands are not available.`.
+> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands won't appear in the output of the help command, and if you attempt to execute an invalid command you get the error `This version of picotool was compiled without USB support. Some commands are not available.`. The build output message `libUSB is not found - no USB support will be built` also appears in the build logs.
 
 ### Linux / macOS
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -8,7 +8,7 @@ git submodule update --init lib/mbedtls
 
 You also need to install `libusb-1.0` if you want to use the USB functionality.
 
-> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands won't appear in the output of the help command, and if you attempt to execute an invalid command you get the error `This version of picotool was compiled without USB support. Some commands are not available.`. The build output message `libUSB is not found - no USB support will be built` also appears in the build logs.
+> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be identified by running `picotool version`, which will state `This version of picotool was compiled without USB support. Some commands are not available.`. Additionally, the unsupported commands won't appear in the output of the help command, and if you attempt to execute an invalid command you will get an error message. The build output message `libUSB is not found - no USB support will be built` also appears in the build logs.
 
 ### Linux / macOS
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -8,7 +8,7 @@ git submodule update --init lib/mbedtls
 
 You also need to install `libusb-1.0` if you want to use the USB functionality.
 
-> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands won't appear in the output of the help command. The build output message 'libUSB is not found - no USB support will be built' also appears in the build logs.
+> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands won't appear in the output of the help command. The build output message `libUSB is not found - no USB support will be built` also appears in the build logs, and if you attempt to execute an invalid command you get the error `This version of picotool was compiled without USB support. Some commands are not available.`.
 
 ### Linux / macOS
 
--- a/main.cpp
+++ b/main.cpp
@@ -705,7 +705,11 @@ struct info_command : public cmd {
     }
 
     string get_doc() const override {
+        #if HAS_LIBUSB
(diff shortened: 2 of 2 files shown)


kilograham (approved):

BUILDING.md
```diff
@@ -8,7 +8,7 @@ git submodule update --init lib/mbedtls
 
 You also need to install `libusb-1.0` if you want to use the USB functionality.
 
-> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands won't appear in the output of the help command. The build output message 'libUSB is not found - no USB support will be built' also appears in the build logs.
+> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands won't appear in the output of the help command. The build output message `libUSB is not found - no USB support will be built` also appears in the build logs, and if you attempt to execute an invalid command you get the error `This version of picotool was compiled without USB support. Some commands are not available.`.
```
lurch: IMHO the addition here ought to be added to the end of he previous sentence (where it's talking about the output of the help command), rather than in the sentence talking about build logs? :thinking:

main.cpp
```diff
@@ -1654,6 +1669,11 @@ int parse(const int argc, char **argv) {
             fos.hanging_indent(0);
             fos.wrap_hard();
             fos << "Use \"picotool help <cmd>\" for more info\n";
+            #if !HAS_LIBUSB
+            if (!help_mode) {
+                fos << "\nThis version of picotool was compiled without USB support. Some commands are not available.\n";
```
lurch: As this identical string appears in 3 separate places, I dunno if i might be worth making it a shared constant?

lurch:
Might also be worth modifying the `get_doc` for the `help_command` ?

will-v-pi:
> Might also be worth modifying the `get_doc` for the `help_command` ?

Do you mean this bit?
```
$ picotool help
PICOTOOL:
    Tool for interacting with RP-series device(s) in BOOTSEL mode, or with an RP-series binary
```
This already changes when you don't have libusb to
```
$ picotool help
PICOTOOL:
    Tool for interacting with an RP-series binary
```

lurch:
> This already changes when you don't have libusb to
> ```
> $ picotool help
> PICOTOOL:
>     Tool for interacting with an RP-series binary
> ```

Yeah, I was thinking that it might be nice to also display the "This version of picotool was compiled without USB support. Some commands are not available." text when running `picotool help`? (Following the "better to be explicit than implicit" principle :wink: )

will-v-pi:
> > This already changes when you don't have libusb to
> > ```
> > $ picotool help
> > PICOTOOL:
> >     Tool for interacting with an RP-series binary
> > ```
> 
> Yeah, I was thinking that it might be nice to also display the "This version of picotool was compiled without USB support. Some commands are not available." text when running `picotool help`? (Following the "better to be explicit than implicit" principle 😉 )

Ah, I'd actually been explicit not to do that (the `if (!help_mode)` line), to avoid contaminating the output of valid commands? With the current change, it only gets printed when you either do something invalid, or with `picotool version` which seemed like a tidier place to put it

main.cpp
```diff
@@ -114,6 +114,10 @@ static const string rp2350_arm_s_family_name = "rp2350-arm-s";
 static const string rp2350_arm_ns_family_name = "rp2350-arm-ns";
 static const string rp2350_riscv_family_name = "rp2350-riscv";
 
+#if !HAS_LIBUSB
+static const string libusb_not_found_message = "\nThis version of picotool was compiled without USB support. Some commands are not available.\n";
```
lurch: Super minor nitpick, but IMHO `libusb_not_found_message` makes it sound like it's looking for libusb at runtime? `built_without_libusb_message` might be slightly better? :shrug:

BUILDING.md
```diff
@@ -8,7 +8,7 @@ git submodule update --init lib/mbedtls
 
 You also need to install `libusb-1.0` if you want to use the USB functionality.
 
-> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands won't appear in the output of the help command. The build output message 'libUSB is not found - no USB support will be built' also appears in the build logs.
+> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be recognized because these commands won't appear in the output of the help command, and if you attempt to execute an invalid command you get the error `This version of picotool was compiled without USB support. Some commands are not available.`. The build output message `libUSB is not found - no USB support will be built` also appears in the build logs.
```
lurch: _Maybe_ it's worth rewording this slightly to mention that `picotool version` is now the easiest way to determine whether USB support is built in or not?

lurch (approved):

PR #252, opened 2025-07-14
Check hamming weights of AES key shares

will-v-pi:
This checks that the hamming weights of the shares are all similar. Currently set to a max difference of 1 between shares (`max_weight_delta`), but that can be changed before merging depending on what distance is actually required for security.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -5173,13 +5173,16 @@ bool encrypt_command::execute(device_map &devices) {
         }
     }
 
+    uint8_t max_weight_delta = 1;
+
     if (!keyIsShare) {
         // Generate a random key share from 256-bit key
         std::random_device rand{};
         assert(rand.max() - rand.min() >= 256);
         for(int i=0; i < 8; i++) {
             // Regenerate the share word until the hamming weights are close to each other
-            while (true) {
+            bool pass = false;
+            for (int attempt=0; attempt < 100000; attempt++) {
                 for (int j=0; j < 12; j++) {
                     aes_key_share.bytes[i*16 + j] = rand();
                 }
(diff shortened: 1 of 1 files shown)


main.cpp
```diff
@@ -5173,18 +5173,63 @@ bool encrypt_command::execute(device_map &devices) {
...
-        for(int i=0; i < 8; i++) {
-            for (int j=0; j < 12; j++) {
-                aes_key_share.bytes[i*16 + j] = rand();
-            }
-            aes_key_share.words[i*4 + 3] = aes_key.words[i]
-                                        ^ aes_key_share.words[i*4]
-                                        ^ aes_key_share.words[i*4 + 1]
-                                        ^ aes_key_share.words[i*4 + 2];
+        // Regenerate the shares until the hamming weights are close to each other
+        while (true) {
```
kilograham: I worry about whether this is guaranteed to terminate (may change depending on other comments)... how many loops does it do for you on average; I think we should limit and fail if it takes too many (or too long)

main.cpp
```diff
@@ -5173,18 +5173,63 @@ bool encrypt_command::execute(device_map &devices) {
...
+                                            ^ aes_key_share.words[i*4]
+                                            ^ aes_key_share.words[i*4 + 1]
+                                            ^ aes_key_share.words[i*4 + 2];
+            }
+
+            uint8_t max_weight = 0;
+            uint8_t min_weight = 128;
+            for (int i=0; i < 4; i++) {
+                uint8_t weight = 0;
+                for (int j=0; j < 4; j++) {
```
kilograham: I assume this should be < 8

kilograham: that said, i don't think we're checking the right thing anyway.

i think:

1. we want the hamming weigths of each of the shares to be about the same (in OTP row - 16 bit - sized chunks). i.e. the first 16 bits of share0, 1, 2, 3 should have similar hamming wieghts
2. In general i think the hamming weights should be about 50% (i.e. 7,8,9 maybe)

Note that perhaps this is likely in general if the random number generator is good

will-v-pi: I've switched it over to checking in 16-bit chunks, and can change it to check for hamming weights in range 7-9 instead if that's what we want (rather than just checking hamming weights of different shares are within 1)

main.cpp
will-v-pi: I've changed it to a 100,000 limit - generally it seems to terminate within 100-1500 attempts, so 100,000 should be plenty

kilograham (approved):

Issue #251, opened 2025-07-10
Cannot build on windows with libusb


JohnAZoidberg:
I tried to do it myself and then I noticed the GitHub action CI also can't do it:

```
-- Could NOT find LIBUSB (missing: LIBUSB_LIBRARIES LIBUSB_INCLUDE_DIR) 
libUSB is not found - no USB support will be built
```

In the run of "[Build & Test (windows-latest, Ninja, mbedtls, libusb, compile)](https://github.com/raspberrypi/picotool/actions/runs/16102733671/job/45433976581#logs)"

lurch:
Well that's unexpected! @will-v-pi I guess the CI-test should check that in the libusb-included builds, the `picotool help` output lists the `load` command as being available?

will-v-pi:
What version of libusb are you downloading, and did you definitely set the `LIBUSB_ROOT` environment variable, or pass `-DLIBUSB_ROOT=/path/to/libusb` to CMake?

I've just tested with the latest (https://github.com/libusb/libusb/releases/download/v1.0.29/libusb-1.0.29.7z) and it's working fine for me - the issue with the action was not setting the environment variable correctly, which I'm fixing now

JohnAZoidberg:
Thanks, I'll try it again.

Could you have the github actions upload the windows binary - then we could easily get the latest build.

lurch:
If you're unable to build `picotool` yourself, I believe the recommendation is to download it from https://github.com/raspberrypi/pico-sdk-tools/releases

will-v-pi:
Closing as the action has been fixed, and there have been no further reports of this issue

PR #253, opened 2025-07-15
Fix Windows build and CI with libusb

will-v-pi:
Fix Windows build with MSVC, and fix the libusb inclusion in the Windows CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,6 +76,13 @@ jobs:
           picotool info -a blink.uf2
           picotool info -a hello_world.uf2
           picotool info -a flash_nuke.uf2
+      
+      - name: Test libusb compiled
+        if: matrix.libusb == 'libusb'
+        run: |
+          picotool help load
+          picotool help save
+          picotool help erase
 
   test-examples:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -31,10 +31,15 @@ jobs:
           choco install -y .github/workflows/choco_packages.config
           curl -L https://github.com/libusb/libusb/releases/download/v1.0.27/libusb-1.0.27.7z -o libusb.7z
           7z x libusb.7z -olibusb
+          ls libusb
```
lurch, line 34: Was this just a debugging line, which can now be removed?

lurch:
This addresses part of the problem identified in #251

.github/workflows/test.yml
will-v-pi: It’s a debugging line, but leaving it in to aid with future debugging (the output from the job is quite verbose anyway)

ndabas: You could use [output grouping](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#grouping-log-lines) to make it a bit more organized if you wanted.

PR #254, opened 2025-07-17
Small file fixes

will-v-pi:
There are currently a couple of bugs when loading/saving/verifying small BIN files

- <100 byte files cause an infinite loop - fix is to round up size/100
- <256 byte files cause an unmapped exception when no family ID is specified - fix is to allow zero fill in RP2040 checksum

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -339,7 +339,7 @@ private:
 // Calculate chunk size for load/save/verify
 // Returns size/100 rounded up to FLASH_SECTOR_ERASE_SIZE
 uint32_t calculate_chunk_size(uint32_t size) {
-    return ((size/100 + FLASH_SECTOR_ERASE_SIZE - 1) & ~(FLASH_SECTOR_ERASE_SIZE - 1));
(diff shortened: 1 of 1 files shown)


kilograham (approved):

main.cpp
```diff
@@ -339,7 +339,7 @@ template <typename T> struct range_map {
 // Calculate chunk size for load/save/verify
 // Returns size/100 rounded up to FLASH_SECTOR_ERASE_SIZE
 uint32_t calculate_chunk_size(uint32_t size) {
-    return ((size/100 + FLASH_SECTOR_ERASE_SIZE - 1) & ~(FLASH_SECTOR_ERASE_SIZE - 1));
+    return (((size + (100 - 1))/100 + FLASH_SECTOR_ERASE_SIZE - 1) & ~(FLASH_SECTOR_ERASE_SIZE - 1));
```
lurch, line 342: Only a minor nitpick, but this seems to be using two different methods for "rounding up" ?

will-v-pi: One is rounding up and dividing not by a power of 2, the other is just rounding up to a power of 2 without dividing

lurch: Clever!

PR #258, opened 2025-07-18
Update .gitignore to exclude subbuild

BrooklynLJ:
Adding a gitignore entry for files generated during build

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 build/*
 bazel-*
+picotool-subbuild/*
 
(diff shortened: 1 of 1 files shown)


lurch:
Where is this `picotool-subbuild` name coming from? I can't find it anywhere on either the `master` or `develop` branches of this repo?
(Note also, that PRs need to target the `develop` branch)

will-v-pi:
Yes, this is not necessary - the `picotool-subbuild` directory is created when the SDK auto-builds picotool, and is not placed inside the `picotool` tree so this `.gitignore` change is not needed. 

See also #221

PR #263, opened 2025-07-22
Improve forced commands for RP2040 and custom vid/pid

### LINKED ISSUE #260: Forced commands actually do work with RP2040 on Windows ... with some restrictions
Hello,

I see that the forced commands are disabled on Windows.
` #if defined(_WIN32)`
`{`
`  struct libusb_device_descriptor desc;`
`  libusb_get_device_descriptor(to_reboot, &desc);`
`  if (desc.idProduct == PRODUCT_ID_RP2040_STDIO_USB) {`
`    fail(ERROR_NOT_POSSIBLE,`
`    "Forced commands do not work with RP2040 on Windows - you can force reboot into BOOTSEL mode via 'picotool reboot -f -u' instead.");`
`  }`
`}`
`#endif`

But they can work in two cases. Either by binding the device to WinUSB with Zadig when it is in BOOTSEL mode with the MSC disabled. But in this case the MSC won't work anymore (the drive won't show up).

Or if you don't mask the MSC when you reboot the device.
Replacing the following line in main.cpp:
`reboot_device(to_reboot, to_reboot_handle, true, 1);`
by
`reboot_device(to_reboot, to_reboot_handle, true, 0);`

I don't know how to have the device working with picotool with AND without the MSC disabled.

### LINKED ISSUE #261: BUGFIX Forced commands don't work with specified VID/PID
When a forced command is asked, we need to unset the specified USB VID/PID to be able to find the device in BOOTSEL mode (which uses Raspberry VID/PID).
The following lines can be added:
`settings.vid = -1;`
`settings.pid = -1;`

before or after

`reboot_device(to_reboot, to_reboot_handle, true, 1);`

in main.cpp.

Combined with #260 I am able to re-program my RP2040 USB device running any USB class on Windows without debug probe or BOOTSEL button with a single command line.
`picotool load -x path\to\project.elf --vid YOUR_VID --pid YOUR_PID -f`

will-v-pi:
Allow forced commands with RP2040 on Windows - they will not deactivate the mass storage device, to ensure Zadig still works

Also fix forced commands with custom vid/pid (as they will change in bootsel mode), by removing vid/pid filters after reboot and just filtering on serial number

Fixes #260 and #261

### DIFF
--- a/README.md
+++ b/README.md
@@ -1660,6 +1660,7 @@ to unlock our lockout of USB MSD writes (we have turned them off so the user doe
 `picotool info` again will unlock it properly the next time (or you can reboot the device).
 
 ### Zadig
+
 To communicate with RP2040 in BOOTSEL mode on Windows, you will need to install a driver. To do this, download and run [Zadig](http://zadig.akeo.ie), select `RP2 Boot (Interface 1)` from the dropdown box and select `WinUSB` as the driver, and click on the "Install Driver" button. Wait for the installation to complete - this may take a few minutes.
 
 This is only required for RP2040.
--- a/README.md
+++ b/README.md
@@ -1658,3 +1658,8 @@ this requirement (see the [hello_usb](https://github.com/raspberrypi/pico-exampl
 If you ctrl+c out of the middle of a long operation, then libusb seems to get a bit confused, which means we aren't able
 to unlock our lockout of USB MSD writes (we have turned them off so the user doesn't step on their own toes). Simply running
 `picotool info` again will unlock it properly the next time (or you can reboot the device).
+
+### Zadig
+To communicate with RP2040 in BOOTSEL mode on Windows, you will need to install a driver. To do this, download and run [Zadig](http://zadig.akeo.ie), select `RP2 Boot (Interface 1)` from the dropdown box and select `WinUSB` as the driver, and click on the "Install Driver" button. Wait for the installation to complete - this may take a few minutes.
+
+This is only required for RP2040.
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ TARGET SELECTION:
         --ser <ser>
             Filter by serial number
         --rp2040
-            Use RP2040 workarounds on Windows when using custom vid/pid (ignored on other platforms)
+            Use RP2040-specific workarounds when using a custom vid/pid on Windows (ignored on other platforms)
         -f, --force
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
@@ -224,7 +224,7 @@ TARGET SELECTION:
         --ser <ser>
             Filter by serial number
         --rp2040
-            Use RP2040 workarounds on Windows when using custom vid/pid (ignored on other platforms)
+            Use RP2040-specific workarounds when using a custom vid/pid on Windows (ignored on other platforms)
         -f, --force
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
@@ -327,7 +327,7 @@ OPTIONS:
         --ser <ser>
             Filter by serial number
         --rp2040
-            Use RP2040 workarounds on Windows when using custom vid/pid (ignored on other platforms)
+            Use RP2040-specific workarounds when using a custom vid/pid on Windows (ignored on other platforms)
         -f, --force
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
@@ -395,7 +395,7 @@ OPTIONS:
         --ser <ser>
             Filter by serial number
         --rp2040
-            Use RP2040 workarounds on Windows when using custom vid/pid (ignored on other platforms)
+            Use RP2040-specific workarounds when using a custom vid/pid on Windows (ignored on other platforms)
         -f, --force
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
@@ -471,7 +471,7 @@ OPTIONS:
         --ser <ser>
             Filter by serial number
         --rp2040
-            Use RP2040 workarounds on Windows when using custom vid/pid (ignored on other platforms)
+            Use RP2040-specific workarounds when using a custom vid/pid on Windows (ignored on other platforms)
         -f, --force
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
@@ -522,7 +522,7 @@ OPTIONS:
         --ser <ser>
             Filter by serial number
         --rp2040
-            Use RP2040 workarounds on Windows when using custom vid/pid (ignored on other platforms)
+            Use RP2040-specific workarounds when using a custom vid/pid on Windows (ignored on other platforms)
         -f, --force
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
@@ -737,7 +737,7 @@ OPTIONS:
         --ser <ser>
             Filter by serial number
         --rp2040
-            Use RP2040 workarounds on Windows when using custom vid/pid (ignored on other platforms)
+            Use RP2040-specific workarounds when using a custom vid/pid on Windows (ignored on other platforms)
         -f, --force
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
@@ -896,7 +896,7 @@ OPTIONS:
         --ser <ser>
             Filter by serial number
         --rp2040
-            Use RP2040 workarounds on Windows when using custom vid/pid (ignored on other platforms)
+            Use RP2040-specific workarounds when using a custom vid/pid on Windows (ignored on other platforms)
         -f, --force
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
@@ -997,7 +997,7 @@ TARGET SELECTION:
         --ser <ser>
             Filter by serial number
         --rp2040
-            Use RP2040 workarounds on Windows when using custom vid/pid (ignored on other platforms)
+            Use RP2040-specific workarounds when using a custom vid/pid on Windows (ignored on other platforms)
         -f, --force
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be rebooted back to application mode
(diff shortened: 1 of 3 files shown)


main.cpp
```diff
@@ -8762,6 +8772,13 @@ int main(int argc, char **argv) {
                         // again is to assume it has the same serial number.
                         settings.address = -1;
                         settings.bus = -1;
+                        if (settings.pid != -1 || settings.vid != -1) {
+                            // also skip vid/pid filtering, as that should change in BOOTSEL mode, and could be white-labelled on RP2350
+                            settings.pid = -1;
```
lurch: Seems slightly confusing that the comment-line talks about "vid/pid filtering", but the code-line only mentions `settings.pid`?

main.cpp
```diff
@@ -8762,6 +8772,13 @@ int main(int argc, char **argv) {
...
+                            // still filter for rpi vid/pid if we don't have a serial number, as that is an RP2040 running a no_flash binary, so will
+                            // have a standard rpi vid/pid in BOOTSEL mode
+                            settings.vid = settings.ser.empty() ? -1 : 0;   // 0 means skip vid/pid filtering entirely, -1 means filter for rpi vid/pid
```
lurch: Perhaps something like this is slightly less confusing? :man_shrugging: 
```suggestion
                            if (settings.ser.empty()) {
                                // still filter for rpi vid/pid if we don't have a serial number, as that is an RP2040 running a no_flash binary, so will
                                // have a standard rpi vid/pid in BOOTSEL mode
                                settings.vid = -1; // -1 means filter for rpi vid/pid
                                settings.pid = -1;
                            } else {
                                // also skip vid/pid filtering, as that should change in BOOTSEL mode, and could be white-labelled on RP2350
                                settings.vid = 0; // 0 means skip vid/pid filtering entirely
                                settings.pid = -1;
                            }
```

lurch:
I _think_ the "Use RP2040 workarounds on Windows when using custom vid/pid (ignored on other platforms)" stuff is assuming that if you're running a UF2 on an RP2350 with a custom VID/PID, then you'll have **also** modified the RP2350's OTP to report the same VID/PID in BOOTSEL mode? But I guess that _might_ not always be the case? :thinking:

will-v-pi:
> I _think_ the "Use RP2040 workarounds on Windows when using custom vid/pid (ignored on other platforms)" stuff is assuming that if you're running a UF2 on an RP2350 with a custom VID/PID, then you'll have **also** modified the RP2350's OTP to report the same VID/PID in BOOTSEL mode? But I guess that _might_ not always be the case? 🤔

Nope, the workaround is just to not disable the MSD, nothing to do with the actual vid/pid - if you have RP2040 stdio_usb vid/pid it autodetects that and does it automatically, but if you're using custom vid/pid then you need to tell it you have an RP2040 as it can't detect it from the vid/pid

lurch:
> > I _think_ the "Use RP2040 workarounds on Windows when using custom vid/pid (ignored on other platforms)" stuff is assuming that if you're running a UF2 on an RP2350 with a custom VID/PID, then you'll have **also** modified the RP2350's OTP to report the same VID/PID in BOOTSEL mode? But I guess that _might_ not always be the case? 🤔
> 
> Nope, the workaround is just to not disable the MSD, nothing to do with the actual vid/pid - if you have RP2040 stdio_usb vid/pid it autodetects that and does it automatically, but if you're using custom vid/pid then you need to tell it you have an RP2040 as it can't detect it from the vid/pid

Okay, and how is the situation different on RP2350?

will-v-pi:
> Okay, and how is the situation different on RP2350?

RP2350 bootrom has the correct MS_OS_20 USB descriptors so works on Windows out of the box - whereas RP2040 doesn’t so requires drivers installed by Zadig, which only work properly when the RP2040 in bootsel mode has the MSD enabled

lurch:
> RP2350 bootrom has the correct MS_OS_20 USB descriptors so works on Windows out of the box - whereas RP2040 doesn’t so requires drivers installed by Zadig, which only work properly when the RP2040 in bootsel mode has the MSD enabled

Ahhhh! I knew it'd probably be something obvious that I'd forgotten :joy:  Thanks for the explanation.

main.cpp
```diff
@@ -593,6 +594,7 @@ auto device_selection =
         (option("--vid") & integer("vid").set(settings.vid).if_missing([] { return "missing vid"; })) % "Filter by vendor id" +
         (option("--pid") & integer("pid").set(settings.pid)) % "Filter by product id" +
         (option("--ser") & value("ser").set(settings.ser)) % "Filter by serial number"
+        + option("--rp2040").set(settings.force_rp2040) % "Use RP2040 workarounds on Windows when using custom vid/pid (ignored on other platforms)"
```
```suggestion
        + option("--rp2040").set(settings.force_rp2040) % "Use RP2040-specific workarounds when using a custom vid/pid on Windows (ignored on other platforms)"
```
? (it niggled me slightly that "on other platforms" was so far away from "Windows", that it wasn't entirely clear what "other platforms" was referring to :wink: )

main.cpp
will-v-pi: Have resolved this with something similar to this

README.md
```diff
@@ -1630,3 +1658,8 @@ this requirement (see the [hello_usb](https://github.com/raspberrypi/pico-exampl
 If you ctrl+c out of the middle of a long operation, then libusb seems to get a bit confused, which means we aren't able
 to unlock our lockout of USB MSD writes (we have turned them off so the user doesn't step on their own toes). Simply running
 `picotool info` again will unlock it properly the next time (or you can reboot the device).
+
+### Zadig
+To communicate with RP2040 in BOOTSEL mode on Windows, you will need to install a driver. To do this, download and run [Zadig](http://zadig.akeo.ie), select `RP2 Boot (Interface 1)` from the dropdown box and select `WinUSB` as the driver, and click on the "Install Driver" button. Wait for the installation to complete - this may take a few minutes.
```
lurch, line 1730:
```suggestion
### Zadig

To communicate with RP2040 in BOOTSEL mode on Windows, you will need to install a driver. To do this, download and run [Zadig](http://zadig.akeo.ie), select `RP2 Boot (Interface 1)` from the dropdown box and select `WinUSB` as the driver, and click on the "Install Driver" button. Wait for the installation to complete - this may take a few minutes.
```
(minor niggle)

kilograham (approved):

main.cpp
```diff
@@ -8703,6 +8705,13 @@ int main(int argc, char **argv) {
                     if (result != dr_error) {
                         devices[result].emplace_back(std::make_tuple(chip, *dev, handle));
                     }
+
+                    if (settings.vid == 0 && !settings.ser.empty() && !devices[dr_vidpid_bootrom_ok].empty()) {
+                        // Searching with no vid/pid filtering (ie opening all devices) can cause issues, so stop
```
kilograham: i have no idea what this means - maybe just improve the comment a  bit

main.cpp
```diff
@@ -8797,13 +8806,14 @@ int main(int argc, char **argv) {
...
+                            unsigned int disable_mask = 1;  // disable MSC interface
     #if defined(_WIN32)
                             {
                                 struct libusb_device_descriptor desc;
                                 libusb_get_device_descriptor(to_reboot, &desc);
-                                if (desc.idProduct == PRODUCT_ID_RP2040_STDIO_USB) {
-                                    fail(ERROR_NOT_POSSIBLE,
-                                        "Forced commands do not work with RP2040 on Windows - you can force reboot into BOOTSEL mode via 'picotool reboot -f -u' instead.");
+                                if (desc.idProduct == PRODUCT_ID_RP2040_STDIO_USB || settings.force_rp2040) {
+                                    disable_mask = 0;   // enable MSC interface so Zadig works correctly
```
kilograham: again - not much context for why this is the case - again maybe improve comment

main.cpp
```diff
@@ -8841,6 +8851,18 @@ int main(int argc, char **argv) {
                         // again is to assume it has the same serial number.
                         settings.address = -1;
                         settings.bus = -1;
+                        if (settings.pid != -1 || settings.vid != -1) {
+                            // vid/pid filtering was enabled, but should change in BOOTSEL mode, so needs to be disabled
```
kilograham: should -> may ?

main.cpp
```diff
@@ -610,6 +611,7 @@ auto device_selection =
         (option("--vid") & integer("vid").set(settings.vid).if_missing([] { return "missing vid"; })) % "Filter by vendor id" +
         (option("--pid") & integer("pid").set(settings.pid)) % "Filter by product id" +
         (option("--ser") & value("ser").set(settings.ser)) % "Filter by serial number"
+        + option("--rp2040").set(settings.force_rp2040) % "Use RP2040-specific workarounds when using a custom vid/pid on Windows (ignored on other platforms)"
```
kilograham: im still not quite sure I'd know what this does - basically this is a "you must pass this when using a custom VID/PID on windows and want to do a force reboot"?

kilograham (approved):
I'm approving pending some comment/help text clarification - then go ahead and merge

main.cpp
kilograham: actually force_rp2040 is a confusing name for a variable anyway, and "--rp2040" should probably be the name if and only if it means "please assume the device is an rp2040"

kilograham (approved):

main.cpp
will-v-pi: I think "please assume the device might be an rp2040" is what it means - for SDK stdio_usb it can tell from the PID so you don’t need to pass it, but if you have a custom VID/PID you need to pass `--rp2040` to enable the workarounds.

If you pass `--rp2040` when using an RP2350 on Windows, it will still work fine, but it will keep the mass storage interface enabled in bootsel mode.

Maybe this is more explicit:
```suggestion
        + option("--rp2040").set(settings.force_rp2040) % "Assume the device might be an RP2040 - this is only required when using a custom vid/pid with an RP2040 on Windows, to enable the RP2040 Windows workarounds, and is ignored on other operating systems"
```

main.cpp
```suggestion
                            // vid/pid filtering was enabled, but may change in BOOTSEL mode, so needs to be disabled
```

main.cpp
will-v-pi: Here’s a more detailed explanation
```suggestion
                                    // the Zadig driver should be setup for the device in BOOTSEL mode with no interfaces disabled,
                                    // as there are none disabled when you plug it in while holding the BOOTSEL button
                                    disable_mask = 0;
```

main.cpp
kilograham: yeah but i dont think the user knows what the workarounds are, so its basically "this is required when using RP2040 with a custom vid/pid on Windows to workaround a Windows problem and is ignored on other platforms" (IDK if workaround a windows problem is true)

lurch: Also, "Assume the device is an RP2040" might be better wording than "Assume the device might be an RP2040" ?

Following on from Graham's suggestion, maybe "to enable the RP2040 Windows workarounds" could be removed entirely, or perhaps replaced with "to enable RP2040-specific workarounds" ?

main.cpp
lurch: Maybe "there are none disabled" -> "all the interfaces are enabled"?

main.cpp
will-v-pi: How about this - I think "operating systems" is clearer than "platforms", as it doesn't depend on the pico_platform
```suggestion
        + option("--rp2040").set(settings.force_rp2040) % "Assume the device is an RP2040 - this is only required when using a custom vid/pid with an RP2040 on Windows, and is ignored on other operating systems"
```

main.cpp
```suggestion
                                    // the Zadig driver should be setup for the device in BOOTSEL mode with no interfaces disabled,
                                    // as all the interfaces are enabled when you plug it in while holding the BOOTSEL button
                                    disable_mask = 0;
```

lurch: Minor tweak, but maybe "while holding the BOOTSEL button" -> "while holding down the BOOTSEL button" ?

main.cpp
will-v-pi: I've added more detail to the comment

PR #267, opened 2025-08-06
Support for RP2350 A2/A3/A4 along with code improvements

kilograham:
* replaced the ocean of one-off IFs with a more OO abstraction


lurch:
~~CI seems very unhappy~~
EDIT: fixed

main.cpp
```diff
@@ -1983,24 +1999,83 @@ uint32_t bootrom_table_lookup_rp2350(memory_access& access, uint16_t tag, uint16
...
+    }
+    return models::unknown;
+}
+
+static inline bool is_transfer_aligned(uint32_t addr, const model_t& model) {
+    enum memory_type t = get_memory_type(addr, model);
+    return t != invalid && !(t == flash && addr & (PAGE_SIZE-1));
+}
+
+// this must be called after the right model is set on raw_access whihc is why it isn't
```
```suggestion
// this must be called after the right model is set on raw_access which is why it isn't
```

main.cpp
```diff
@@ -1867,25 +1887,21 @@ struct memory_access {
     }
 
     template <typename T> void read_into_vector(uint32_t addr, unsigned int count, vector<T> &v, bool zero_fill = false) {
-        v.clear();
+            v.clear();
```
lurch: Undesired whitespace change?

picoboot_connection/picoboot_connection_cxx.h
```diff
@@ -28,7 +27,7 @@ namespace picoboot {
     };
 
     struct connection {
-        explicit connection(libusb_device_handle *device, model_t model, bool exclusive = true) : device(device), model(model), exclusive(exclusive) {
+        explicit connection(libusb_device_handle *device, chip_t chip, bool exclusive = true) : device(device), exclusive(exclusive) {
```
lurch: It _looks_ like the `chip` parameter isn't actually used here?

README.md
```diff
@@ -847,6 +847,9 @@ OPTIONS:
     UF2 Family options
         <family_id>
             family ID for UF2
+    Platform options
+        <platform>
+            optional platform for memory verification (eg rp2040, rp2350)
```
lurch, line 852: I dunno if this is a stupid question, but is it possible to infer the `platform` from the `family_id` ?

will-v-pi: Not always (eg for `absolute` or `data` IDs) - this option allows you to specify it explicitly, if not specified then it does the usual search for a block loop/checksum (see `get_access_model`)

main.cpp
```suggestion
        v.clear();
```

picoboot_connection/picoboot_connection_cxx.h
will-v-pi: Have removed it

main.cpp
```diff
@@ -1983,24 +1999,83 @@ uint32_t bootrom_table_lookup_rp2350(memory_access& access, uint16_t tag, uint16
...
     }
+    return chip_revision;
 }
 #if HAS_LIBUSB
 struct picoboot_memory_access : public memory_access {
-    model_t model = unknown; // must be initialized to something up front as it is referenced before it is set to its final value
     explicit picoboot_memory_access(picoboot::connection &connection) : connection(connection) {
-        model = get_model(*this);
+        // todo remove me temp hack so we can read a larger amount of ROM
+        model = std::make_shared<model_info>(unknown, "hack", 0x10000);
```
will-v-pi: Not required, as `determine_model` only reads `BOOTROM_MAGIC_ADDR` and `BOOTROM_MAGIC_ADDR + 4`, which is allowed by unknown model

```suggestion
```

main.cpp
```diff
@@ -6263,7 +6331,6 @@ bool partition_info_command::execute(device_map &devices) {
 
     if (has_pt) {
         picoboot_memory_access raw_access(con);
```
will-v-pi: Doesn't need this, as it's already created on 6299

```suggestion
```

main.cpp
```diff
@@ -7393,9 +7462,18 @@ static void check_otp_write_error(picoboot::command_failure &e, bool ecc) {
...
+
+uint8_t otp_cmd_max_bits(void) {
+    return settings_select_ecc() ? 16 : 24;
+}
+
 typedef std::function<void(uint8_t *buffer, uint32_t len, picoboot_otp_cmd &otp_cmd)> otp_read_func_t;
 typedef std::function<void(uint8_t *buffer, uint32_t len, picoboot_otp_cmd &otp_cmd)> otp_write_func_t;
-void process_otp_json(json &otp_json, otp_read_func_t read_func, otp_write_func_t write_func) {
+void process_otp_json(json &otp_json, model_t model, otp_read_func_t read_func, otp_write_func_t write_func) {
+    int raw_max_bits = otp_cmd_max_bits();
```
will-v-pi: Technically this should always be the maximum, ignoring whether ECC is set, because the JSON file specifies ECC or not for each field so ignores the `settings.otp.ecc` and `settings.otp.raw` options.

Simple fix is
```suggestion
    int raw_max_bits = 24
```
but could add an `otp_raw_max_bits(void)` function?

main.cpp
```diff
@@ -8783,27 +8865,15 @@ int main(int argc, char **argv) {
...
     } catch (picoboot::command_failure& e) {
-        // todo rp2350/rp2040
-        string device = "RP-series";
-        if (selected_model == rp2040) {
-            device = "RP2040";
-        } else if (selected_model == rp2350) {
-            device = "RP2350";
-        }
+        std::string device = chip_name(selected_chip);
         std::cout << "ERROR: The " << device << " device returned an error: " << e.what() << "\n";
```
will-v-pi: Can be simplified

```suggestion
        std::cout << "ERROR: The " << chip_name(selected_chip) << " device returned an error: " << e.what() << "\n";
```

main.cpp
```diff
@@ -8783,27 +8865,15 @@ int main(int argc, char **argv) {
...
     } catch (picoboot::connection_error&) {
-        // todo rp2350/rp2040
-        string device = "RP-series";
-        if (selected_model == rp2040) {
-            device = "RP2040";
-        } else if (selected_model == rp2350) {
-            device = "RP2350";
-        }
+        std::string device = chip_name(selected_chip);
         std::cout << "ERROR: Communication with " << device << " device failed\n";
```
will-v-pi:
```suggestion
        std::cout << "ERROR: Communication with " << chip_name(selected_chip) << " device failed\n";
```

will-v-pi (approved):
Tested and all LGTM

Issue #270, opened 2025-08-10
riscv firmware loaded but arm core selected + some automation questions


cederom:
Hello world :-)

* We are working on Apache NuttX RTOS port to RISC-V core on RP2350 (https://github.com/apache/nuttx/pull/16424).
* UF2 file is generated with picotool and has all properties  set to run on RISC-V:
  ```
  % picotool info -a nuttx.uf2
  File nuttx.uf2 family ID 'rp2350-riscv':
  
  Program Information
   target chip:         RP2350
   image type:          RISC-V
  
  Fixed Pin Information
   none
  
  Build Information
   none
  
  Metadata Block 1
   address:             0x10000000
   next block address:  0x10000000
   block type:          image def
   target chip:         RP2350
   image type:          RISC-V
   entry point:         EP 0x10000020, SP 0x00000000
  
  Metadata Block 2
   address:             0x10000000
   next block address:  0x10000000
   block type:          image def
   target chip:         RP2350
   image type:          RISC-V
   entry point:         EP 0x10000020, SP 0x00000000
  ```
* This image is recognised and flashed to `rp2350-riscv` partition:
  ```
  % picotool load nuttx.uf2
  Family ID 'rp2350-riscv' can be downloaded in absolute space:
    00000000->02000000
  Loading into Flash:   [==============================]  100%
  ```
* After `picotool reboot` device reboots into riscv application:
  ```
  % picotool reboot
  % cu -l /dev/cuaU0 -s 115200
  NuttShell (NSH) NuttX-10.4.0
  nsh> uname -a
  NuttX 10.4.0 35fdf7e495 Aug 10 2025 21:23:16 risc-v raspberrypi-pico-2-rv
  nsh> getprime
  Set thread priority to 10
  Set thread policy to SCHED_RR
  Start thread #0
  thread #0 started, looking for primes < 10000, doing 10 run(s)
  thread #0 finished, found 1230 primes, last one was 9973
  Done
  getprime took 4274 msec
  ```
* The problem is after detach/reattach device does not start RISCV firmware and the ARM cpu seems to be selected as default:
  ```
  % picotool info -a
  Program Information
   target chip:            RP2350
   image type:             RISC-V
  
  Fixed Pin Information
   none
  
  Build Information
   none
  
  Metadata Block 1
   address:                0x10000000
   next block address:     0x10000000
   block type:             image def
   target chip:            RP2350
   image type:             RISC-V
   entry point:            EP 0x10000020, SP 0x00000000
[...]

PR #271, opened 2025-08-12
Prevent printing metadata block twice

will-v-pi:
Prevent printing metadata block twice in `picotool info`, when only one block is present in the block loop

Noticed in #270

Before:
```
Metadata Block 1
 address:                0x10000014
 next block address:     0x10000014
 block type:             image def
 target chip:            RP2350
 image type:             RISC-V
 entry point:            EP 0x10000036, SP 0x20082000

Metadata Block 2
 address:                0x10000014
 next block address:     0x10000014
 block type:             image def
 target chip:            RP2350
 image type:             RISC-V
 entry point:            EP 0x10000036, SP 0x20082000
```

After:
```
Metadata Block 1
 address:             0x10000014
 next block address:  0x10000014
 block type:          image def
 target chip:         RP2350
 image type:          RISC-V
 entry point:         EP 0x10000036, SP 0x20082000
```

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -3697,15 +3697,15 @@ void info_guts(memory_access &raw_access, void *con) {
                 bin = raw_access.read_vector<uint8_t>(raw_access.get_binary_start(), read_size, true);
                 std::unique_ptr<block> first_block = find_first_block(bin, raw_access.get_binary_start());
                 if (first_block) {
-                    // verify stuff
-                    auto all_blocks = get_all_blocks(bin, raw_access.get_binary_start(), first_block, more_cb);
-
                     int block_i = 0;
                     select_group(metadata_info[block_i++], true);
                     info_metadata(first_block.get(), true);
-                    for (auto &block : all_blocks) {
-                        select_group(metadata_info[block_i++], true);
-                        info_metadata(block.get(), true);
+                    if (first_block->next_block_rel != 0) { // if there is more than one block
+                        auto all_blocks = get_all_blocks(bin, raw_access.get_binary_start(), first_block, more_cb);
+                        for (auto &block : all_blocks) {
+                            select_group(metadata_info[block_i++], true);
+                            info_metadata(block.get(), true);
+                        }
                     }
                 } else {
                     // This displays that there are no metadata blocks


main.cpp
```diff
@@ -3697,15 +3697,15 @@ void info_guts(memory_access &raw_access, void *con) {
...
-                    auto all_blocks = get_all_blocks(bin, raw_access.get_binary_start(), first_block, more_cb);
-
                     int block_i = 0;
                     select_group(metadata_info[block_i++], true);
                     info_metadata(first_block.get(), true);
-                    for (auto &block : all_blocks) {
-                        select_group(metadata_info[block_i++], true);
-                        info_metadata(block.get(), true);
+                    if (first_block->next_block_rel != 0) { // if there is more than one block
+                        auto all_blocks = get_all_blocks(bin, raw_access.get_binary_start(), first_block, more_cb);
```
lurch, line 3704: Hmmm, does this change imply that if there are two or more blocks, then `get_all_blocks` doesn't return the first block; but if there's only one block, then `get_all_blocks` _does_ return the first block??
Or am I just mis-reading the diff?

lurch:
Maybe it's worth adding a regression-test for this to CI? :thinking:

main.cpp
will-v-pi: Yes, `get_all_blocks` returns all other blocks when there are multiple blocks, but it returns the first block again if there is only one block

lurch: This might be a stupid question (I'm good a those :laughing: ), but I wonder if that suggests that `get_all_blocks` should instead return zero blocks if there is only one block? Seems like it might be more consistent behaviour? :man_shrugging:

kilograham (approved):

main.cpp
will-v-pi: That's probably for another PR, as I'd need to fix some places it's used so that behaviour works

will-v-pi:
> Maybe it's worth adding a regression-test for this to CI? 🤔

That'd probably be a separate PR, to add regression testing to the `info` output (lots of other bits to check too, eg the binary info)

main.cpp
lurch: Fair enough; not being familiar with the code, I didn't realise that you were relying on the existing behaviour in other places.

PR #273, opened 2025-08-22
Add support for extra security bit

will-v-pi:
Set this by default when signing Arm images, in addition to adding a vector table and entry point if they're not present

This bit has no effect on chip revisions <A4, on A4 it requires that the image has both a vector table and an entry point

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -3362,8 +3362,8 @@ void info_guts(memory_access &raw_access, void *con) {
                     info_pair("tbyb", "not bought");
                 }
 
-                if (image_def->extra_security() && verbose_metadata) {
-                    info_pair("extra security", "enabled");
+                if (verbose_metadata) {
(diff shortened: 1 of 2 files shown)


kilograham (approved):

main.cpp
```diff
@@ -3361,6 +3361,10 @@ void info_guts(memory_access &raw_access, void *con) {
                 if (image_def->tbyb()) {
                     info_pair("tbyb", "not bought");
                 }
+
+                if (image_def->extra_security() && verbose_metadata) {
+                    info_pair("extra security", "enabled");
```
lurch: Should it print `info_pair("extra security", "not enabled")` if the flag _hasn't_  been set? :thinking:

will-v-pi: Yes, that's a good idea

PR #276, opened 2025-09-04
add documentation for reboot --diagnostic option

josch:


will-v-pi (approved):
Thanks, LGTM

lurch:
Does the README.md also need to be updated @will-v-pi ?

will-v-pi:
It looks like the reboot command is missing from the README entirely, and it's not being caught by the script because it's defined slightly differently to the others

lurch:
> It looks like the reboot command is missing from the README entirely, and it's not being caught by the script because it's defined slightly differently to the others

Ooops :wink:

josch:
> It looks like the reboot command is missing from the README entirely, and it's not being caught by the script because it's defined slightly differently to the others

What is "the script" if I may ask? :)

I only noticed that the --diagnostic option was missing documentation because my own set of scripts broke due to the missing docs:

 * https://sources.debian.org/src/picotool/2.2.0-a4%2Bdfsg-2/debian/manpages.sh
 * https://sources.debian.org/src/picotool/2.2.0-a4%2Bdfsg-2/debian/help2md.pl

The scripts parse the --help output and turn it into a man page using pandoc. help2man is unfortunately not able to parse the specific format of picotool --help.

lurch:
> What is "the script" if I may ask? :)

See https://github.com/raspberrypi/picotool/blob/master/.github/workflows/check_help_text.yml and https://github.com/raspberrypi/picotool/blob/master/gen_help_txt.sh

josch:
> https://github.com/raspberrypi/picotool/blob/master/gen_help_txt.sh

Ah, good to know that processing main.cpp with regexes is actually not as evil as I thought it would be. I'll add the fact that you are doing the same thing in `gen_help_txt.sh` to my scripts, thank you!

PR #280, opened 2025-09-15
Add reboot command to readme

will-v-pi:
Add a description of the `picotool reboot` command to the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -566,7 +566,40 @@ Partition 1
 
 ```text
 $ picotool help reboot
-    Let CI do it
+REBOOT:
+    Reboot the device
+
+SYNOPSIS:
+    picotool reboot [-a] [-u] [-g <partition>] [-c <cpu>] [device-selection]
+
+OPTIONS:
+    Reboot type
+        -a, --application
+            Reboot back into the application (this is the default)
+        -u, --usb
+            Reboot back into BOOTSEL mode
+        -g <partition>
+            Select diagnostic partition
+        -c <cpu>
+            Select arm | riscv CPU (if possible)
+    Selecting the device to reboot
+        --bus <bus>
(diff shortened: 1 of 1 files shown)


lurch:
Probably also makes sense to update the "checking script" so that this (or other similar commands) can't get overlooked in future? :thinking:

will-v-pi:
> Probably also makes sense to update the "checking script" so that this (or other similar commands) can't get overlooked in future? 🤔

The `reboot` command would need it's own more complex handling added to the script, as the current script checks for the array of all the commands, but the reboot command is created outside of that array

Given it's unlikely there'll be another command like `reboot` which needs this handling, as I've moved to using `load_guts`, `info_guts` etc instead of calling `reboot_cmd->execute`, I think it's fine to not add handling to the script - now the command is in the README, it'll still be updated if the help text changes anyway

lurch:
> now the command is in the README, it'll still be updated if the help text changes anyway

Fair enough :+1:

README.md
```diff
@@ -560,6 +560,48 @@ Partition 1
...
+    picotool reboot [-a] [-u] [-g <partition>] [-c <cpu>] [device-selection]
+
+OPTIONS:
+    Reboot type
+        -a, --application
+            Reboot back into the application (this is the default)
+        -u, --usb
+            Reboot back into BOOTSEL mode
+        -g <partition>
+            Select diagnostic partition
```
lurch, line 582: @kilograham Could we get a brief explanation of what the diagnostic partition is please?

will-v-pi: I will merge this as is for now, but feel free to create a separate issue to better describe what the diagnostic partition is

lurch: Nobody else has asked about it, so I assume it's not important!

PR #281, opened 2025-09-16
Improve docs for picotool load -x

will-v-pi:
Mention that is performs a flash update boot, to tie in with term in the datasheet (used for TBYB and version downgrades)

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -854,7 +854,7 @@ struct load_command : public cmd {
                 option('N', "--no-overwrite-unsafe").set(settings.load.no_overwrite_force) % "When writing flash data, do not overwrite an existing program in flash. If picotool cannot determine the size/presence of the program in flash, the load continues anyway" +
                 option('u', "--update").set(settings.load.update) % "Skip writing flash sectors that already contain identical data" +
                 option('v', "--verify").set(settings.load.verify) % "Verify the data was written correctly" +
-                option('x', "--execute").set(settings.load.execute) % "Perform a flash update boot to execute the downloaded file as a program after the load"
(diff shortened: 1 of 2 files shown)


main.cpp
```diff
@@ -854,7 +854,7 @@ struct load_command : public cmd {
                 option('N', "--no-overwrite-unsafe").set(settings.load.no_overwrite_force) % "When writing flash data, do not overwrite an existing program in flash. If picotool cannot determine the size/presence of the program in flash, the load continues anyway" +
                 option('u', "--update").set(settings.load.update) % "Skip writing flash sectors that already contain identical data" +
                 option('v', "--verify").set(settings.load.verify) % "Verify the data was written correctly" +
-                option('x', "--execute").set(settings.load.execute) % "Attempt to execute the downloaded file as a program after the load"
+                option('x', "--execute").set(settings.load.execute) % "Perform a flash update boot to execute the downloaded file as a program after the load"
```
lurch: Is a "flash update boot" specific to RP2350? (i.e. was the old help-text correct for RP2040?)

andygpz11 (approved):
Thanks, much better.

andygpz11 (approved):
Thanks, much better.

main.cpp
will-v-pi: Technically yes, but I think it'd be safe to call it a flash update boot on both devices (ie on RP2040 there is no distinction between a flash update boot and a normal boot)

Although you remind me that I've missed SRAM boots from this description, so maybe something like

```suggestion
                option('x', "--execute").set(settings.load.execute) % "Perform a bootrom reboot to execute the downloaded file as a program after the load - either a flash update boot for binaries in flash, or a RAM image boot for other binaries "
```

PR #284, opened 2025-10-28
Picotool dependencies solved with Nix

jaumefe:
Added flake to avoid installation of individual dependencies for Linux and MacOS

README file updated to explain how to use nix


will-v-pi:
We don’t plan on merging this as it adds additional flake files we would have to maintain, and the existing installation method works fine

PR #292, opened 2026-02-11
Fix bazel build

will-v-pi:
These rules are no longer included by default

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,3 +1,5 @@
+load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
 load("//bazel:defs.bzl", "otp_header_parse", "picotool_binary_data_header")
 
 package(default_visibility = ["//visibility:public"])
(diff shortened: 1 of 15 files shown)


.github/workflows/bazel_build.yml
```diff
@@ -28,8 +28,8 @@ jobs:
       - name: Fetch latest Pico SDK
         uses: actions/checkout@v4
         with:
-          repository: raspberrypi/pico-sdk
-          ref: develop
+          repository: armandomontanez/pico-sdk
```
kilograham: Was this change intentional?

lurch:
Would the CI checks fare better if this was combined with #293 ?

.github/workflows/bazel_build.yml
will-v-pi: Yes - I’ll remove it once the SDK change is merged, and then mark this PR as ready

will-v-pi: SDK PR is https://github.com/raspberrypi/pico-sdk/pull/2825

will-v-pi:
> Would the CI checks fare better if this was combined with #293 ?

I'm keeping them separate, as they're separate CI jobs - but yes, CI should all work after both these changes

PR #289, opened 2026-01-30
Fix MSVC build problems caused by 'windows.h'

WagnerChristoph:
Fix the problem when building with MSVC due to a macro conflict, arising from including `windows.h` without a preceding `#define NOMINMAX` (as it re-defines `min`/`max` otherwise).  
In this case, [`picoboot_connection.h`](https://github.com/raspberrypi/picotool/blob/master/picoboot_connection/picoboot_connection.h) included the `libusb.h` header unguardedly, which in turn includes `windows.h`.  This was previously fixed with `#undef min/max`, but in [`main.cpp`](https://github.com/raspberrypi/picotool/blob/master/main.cpp) only *after* including [`model.h`](https://github.com/raspberrypi/picotool/blob/master/model/model.h), where the error occurred.
Also added the guard to [`portable_endian.h`](https://github.com/raspberrypi/picotool/blob/master/elf/portable_endian.h), the other place where `windows.h` might be included (through `winsock2.h`), and cleaned up the now obsolete `#undef`s.

Fixes #288.

PR #293, opened 2026-02-11
Fix MSVC build and CI tests

will-v-pi:
Fix MSVC build, and add CI test of the MSVC build

Also fix the CI tests, as the UF2 files may now be ZIP files, so may need unzipping before use


lurch:
I guess this partly overlaps with #289 ?

will-v-pi:
> I guess this partly overlaps with #289 ?

Yes, this fixes that issue the same way we have done for other files - I will merge this one first, as it adds CI tests which can test #289

PR #298, opened 2026-03-06
Add support for pinning XIP SRAM

will-v-pi:
Adds support for pinning XIP SRAM in a load map, and allows use of XIP SRAM in flash binaries

### DIFF
--- a/README.md
+++ b/README.md
@@ -607,7 +607,7 @@ OPTIONS:
 
 `seal` allows you to sign and/or hash a binary to run on RP2350.
 
-By default, it will just add an embedded block to the end containing a relative load map, thus sealing the binary. The `--hash` and `--sign` arguments can be passed to include an hash and a signature; `--major`, `--minor` and `--rollback` can be passed to include a version; and `--clear` or `--pin-xip-sram` can be passed to add load map items indicating clear of SRAM and pin of XIP SRAM.
+By default, it will just add an embedded block to the end containing a relative load map, thus sealing the binary. The `--hash` and `--sign` arguments can be passed to include an hash and a signature; `--major`, `--minor` and `--rollback` can be passed to include a version; and `--clear` or `--pin-xip-sram` can be passed to add load map items indicating whether SRAM should be cleared and whether XIP SRAM should be pinned.
 
 Your signing key must be for the _secp256k1_ curve, in PEM format. You can create a .PEM file with:
 
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ SYNOPSIS:
     picotool config [-s <key> <value>] [-g <group>] <filename> [-t <type>]
     picotool load [--ignore-partitions] [--family <family_id>] [-p <partition>] [-n] [-N] [-u] [-v] [-x] <filename> [-t <type>] [-o
                 <offset>] [device-selection]
-    picotool encrypt [--quiet] [--verbose] [--embed] [--fast-rosc] [--use-mbedtls] [--otp-key-page <page>] [--hash] [--sign] [--clear]
+    picotool encrypt [--quiet] [--verbose] [--embed] [--fast-rosc] [--use-mbedtls] [--otp-key-page <page>] [--hash] [--sign] [--no-clear]
                 [--pin-xip-sram] <infile> [-t <type>] [-o <offset>] <outfile> [-t <type>] <aes_key> <iv_salt> <signing_key> <otp>
     picotool seal [--quiet] [--verbose] [--hash] [--sign] [--clear] [--pin-xip-sram] <infile> [-t <type>] [-o <offset>] <outfile> [-t
                 <type>] <key> <otp> [--major <major>] [--minor <minor>] [--rollback <rollback> [<rows>..]]
@@ -693,7 +693,7 @@ ENCRYPT:
     Encrypt the program.
 
 SYNOPSIS:
-    picotool encrypt [--quiet] [--verbose] [--embed] [--fast-rosc] [--use-mbedtls] [--otp-key-page <page>] [--hash] [--sign] [--clear]
+    picotool encrypt [--quiet] [--verbose] [--embed] [--fast-rosc] [--use-mbedtls] [--otp-key-page <page>] [--hash] [--sign] [--no-clear]
                 [--pin-xip-sram] <infile> [-t <type>] [-o <offset>] <outfile> [-t <type>] <aes_key> <iv_salt> <signing_key> <otp>
 
 OPTIONS:
@@ -724,8 +724,8 @@ OPTIONS:
             Hash the encrypted file
         --sign
             Sign the encrypted file
-        --clear
-            Clear all of SRAM on load
+        --no-clear
+            Don't clear all of SRAM on load
         --pin-xip-sram
             Pin XIP SRAM on load
     File to load from
--- a/README.md
+++ b/README.md
@@ -645,7 +645,7 @@ OPTIONS:
         --sign
             Sign the file
         --clear
-            Clear all of SRAM on load
+            Clear all of main SRAM on load
         --pin-xip-sram
             Pin XIP SRAM on load
     File to load from
@@ -725,7 +725,7 @@ OPTIONS:
         --sign
             Sign the encrypted file
         --no-clear
-            Don't clear all of SRAM on load
+            Don't clear all of main SRAM on load
         --pin-xip-sram
             Pin XIP SRAM on load
     File to load from
--- a/README.md
+++ b/README.md
@@ -16,10 +16,10 @@ SYNOPSIS:
     picotool config [-s <key> <value>] [-g <group>] <filename> [-t <type>]
     picotool load [--ignore-partitions] [--family <family_id>] [-p <partition>] [-n] [-N] [-u] [-v] [-x] <filename> [-t <type>] [-o
                 <offset>] [device-selection]
-    picotool encrypt [--quiet] [--verbose] [--embed] [--fast-rosc] [--use-mbedtls] [--otp-key-page <page>] [--hash] [--sign] <infile> [-t
-                <type>] [-o <offset>] <outfile> [-t <type>] <aes_key> <iv_salt> <signing_key> <otp>
-    picotool seal [--quiet] [--verbose] [--hash] [--sign] [--clear] <infile> [-t <type>] [-o <offset>] <outfile> [-t <type>] <key> <otp>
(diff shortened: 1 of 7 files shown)


bintool/bintool.cpp
```diff
@@ -761,24 +791,41 @@ std::vector<uint8_t> get_lm_hash_data(elf_file *elf, block *new_block, bool clea
...
     std::shared_ptr<load_map_item> load_map = new_block->get_item<load_map_item>();
+    if (check_generic_load_map(load_map, model, pin_xip_sram)) {
+        new_block->items.erase(std::remove(new_block->items.begin(), new_block->items.end(), load_map), new_block->items.end());
+        load_map = nullptr;
+    }
     if (load_map == nullptr) {
-        to_hash.insert(to_hash.begin(), bin.begin(), bin.end());
         std::vector<load_map_item::entry> entries;
         if (clear_sram) {
             // todo gate this clearing of SRAM
```
lurch: Has this `todo` been made redundant by the `clear_sram` flag? Or does the todo mean something else?

lurch:
For people like me who have no idea what "pinning XIP SRAM" is about, maybe it's worth adding a comment pointing to 5.9.3.2.1. in the RP2350 datasheet? :wink:

lurch:
"If, however, you have a non-flash binary with code or data in the XIP RAM address space, then you need to add a special
LOAD_MAP entry to indicate to the bootrom that the XIP contents should be pinned."
Does that imply that you could infer from the binary itself whether the XIP RAM needs to be pinned, rather than requiring an additional CLI flag?

bintool/bintool.cpp
will-v-pi: Yes, it should be this todo instead
```suggestion
            // todo tidy up this way of hashing the uint32_t
```

will-v-pi:
> "If, however, you have a non-flash binary with code or data in the XIP RAM address space, then you need to add a special LOAD_MAP entry to indicate to the bootrom that the XIP contents should be pinned." Does that imply that you could infer from the binary itself whether the XIP RAM needs to be pinned, rather than requiring an additional CLI flag?

Not necessarilly, because you could choose to pin manually only the region you want before loading your code/data into XIP SRAM, rather than having the bootrom pin the whole thing, in order to minimise the flash performance loss that comes from pinning XIP SRAM.

lurch:
> you could choose to pin manually only the region you want before loading your code/data into XIP SRAM, rather than having the bootrom pin the whole thing, in order to minimise the flash performance loss that comes from pinning XIP SRAM.

Okay. So is the `--pin-xip-sram` flag being added here doing the "Pin all of XIP SRAM on load" behaviour?

will-v-pi:
> For people like me who have no idea what "pinning XIP SRAM" is about, maybe it's worth adding a comment pointing to 5.9.3.2.1. in the RP2350 datasheet? 😉

That section of the datasheet is currently incorrect (I have a PR open to fix it), but more generally picotool doesn't currently refer to the datasheet and instead assumes prior knowledge of the bootrom functionality (eg signing and hashing, what a load map is, etc), as that stuff is documented elsewhere

will-v-pi:
> > you could choose to pin manually only the region you want before loading your code/data into XIP SRAM, rather than having the bootrom pin the whole thing, in order to minimise the flash performance loss that comes from pinning XIP SRAM.
> 
> Okay. So is the `--pin-xip-sram` flag being added here doing the "Pin all of XIP SRAM on load" behaviour?

Yes, it's adding a load map entry to ask the bootrom to do that

lurch:
Hmmm, very weird that the "Test Build Examples" jobs have suddenly decided to start failing :confused:  It's complaining about a missing file but I don't think that any of the relevant repos have changed recently?

will-v-pi:
BTStack has been bumped in the SDK but not yet in pico-examples, which is causing the failure

lurch:
> BTStack has been bumped in the SDK but not yet in pico-examples, which is causing the failure

Ahhhh! Thanks :grinning:

bintool/bintool.cpp
```diff
@@ -687,22 +687,52 @@ void hash_andor_sign_block(block *new_block, const public_t public_key, const pr
...
+    if (check_generic_load_map(load_map, model, pin_xip_sram)) {
+        new_block->items.erase(std::remove(new_block->items.begin(), new_block->items.end(), load_map), new_block->items.end());
+        load_map = nullptr;
+    }
     if (load_map == nullptr) {
         std::vector<load_map_item::entry> entries;
         if (clear_sram) {
             // todo tidy up this way of hashing the uint32_t
-            std::vector<uint32_t> sram_size_vec = {SRAM_END_RP2350 - SRAM_START};
+            std::vector<uint32_t> sram_size_vec = {model->sram_end() - model->sram_start()};
```
lurch, line 715: Maybe a stupid question, but does the switch from the `SRAM_END_RP2350` constant to the model-supplied `model->sram_end()` mean that there's now an `#include` somewhere that can be dropped? (which would ensure that e.g. `SRAM_END_RP2350` isn't still unintentionally being used somewhere.)

will-v-pi: No, that constant is still used in model.h when creating the RP2350 model

README.md
```diff
@@ -607,7 +607,7 @@ OPTIONS:
 
 `seal` allows you to sign and/or hash a binary to run on RP2350.
 
-By default, it will just sign the binary, but this can be configured with the `--hash` and `--no-sign` arguments.
+By default, it will just add an embedded block to the end containing a relative load map, thus sealing the binary. The `--hash` and `--sign` arguments can be passed to include an hash and a signature; `--major`, `--minor` and `--rollback` can be passed to include a version; and `--clear` or `--pin-xip-sram` can be passed to add load map items.
```
kilograham: that is confusing `--clear` or `--pin-xip-sram` can be passed to add load map items it might be technically correct, but maybe be explicit (e.g. add load map items indicating clear of RAM and pin of XRAM) but not sure if the user cares about the details

bintool/bintool.cpp
```diff
@@ -687,22 +687,52 @@ void hash_andor_sign_block(block *new_block, const public_t public_key, const pr
 }
 
 
-std::vector<uint8_t> get_lm_hash_data(elf_file *elf, block *new_block, bool clear_sram = false) {
+bool check_generic_load_map(std::shared_ptr<load_map_item> load_map, model_t model, bool &pin_xip_sram) {
```
kilograham: this could be better named i think - also a bit weird to both modify and return a var

README.md
kilograham: i guess you did something more like the first suggestion further down

will-v-pi: I've added the more explicit description

bintool/bintool.cpp
will-v-pi: Renamed to `detect_generic_load_map` - it returns whether one was found, and also sets the var for which one was found (currently only pin_xip_sram, but may be more in future), which works nicely with the code where it's used (eg `get_lm_hash_data`) so it can detect either a generic load map from the SDK, or a command line argument.

README.md
```diff
@@ -607,7 +607,7 @@ OPTIONS:
 
 `seal` allows you to sign and/or hash a binary to run on RP2350.
 
-By default, it will just sign the binary, but this can be configured with the `--hash` and `--no-sign` arguments.
+By default, it will just add an embedded block to the end containing a relative load map, thus sealing the binary. The `--hash` and `--sign` arguments can be passed to include an hash and a signature; `--major`, `--minor` and `--rollback` can be passed to include a version; and `--clear` or `--pin-xip-sram` can be passed to add load map items indicating clear of SRAM and pin of XIP SRAM.
```
lurch: Grammar feels a little odd here - is this better?
```suggestion
By default, it will just add an embedded block to the end containing a relative load map, thus sealing the binary. The `--hash` and `--sign` arguments can be passed to include an hash and a signature; `--major`, `--minor` and `--rollback` can be passed to include a version; and `--clear` or `--pin-xip-sram` can be passed to add load map items indicating whether SRAM should be cleared and whether XIP SRAM should be pinned.
```

README.md
```diff
@@ -646,6 +646,8 @@ OPTIONS:
             Sign the file
         --clear
             Clear all of SRAM on load
+        --pin-xip-sram
+            Pin XIP SRAM on load
```
lurch, line 650: Does "all of SRAM" include "XIP SRAM", or are these distinct things? (Perhaps the wording here can be made clearer?)

will-v-pi: No, they're distinct - I think I can call it "main SRAM" to be clearer

will-v-pi: I've only changed it in those option descriptions, as everywhere else doesn't say "all of" so I think just SRAM is fine

README.md
```diff
@@ -607,7 +607,7 @@ OPTIONS:
 
 `seal` allows you to sign and/or hash a binary to run on RP2350.
 
-By default, it will just sign the binary, but this can be configured with the `--hash` and `--no-sign` arguments.
+By default, it will just add an embedded block to the end containing a relative load map, thus sealing the binary. The `--hash` and `--sign` arguments can be passed to include an hash and a signature; `--major`, `--minor` and `--rollback` can be passed to include a version; and `--clear` or `--pin-xip-sram` can be passed to add load map items indicating whether SRAM should be cleared and whether XIP SRAM should be pinned.
```
lurch, line 610:
```suggestion
By default, it will just add an embedded block to the end containing a relative load map, thus sealing the binary. The `--hash` and `--sign` arguments can be passed to include an hash and a signature; `--major`, `--minor` and `--rollback` can be passed to include a version; and `--clear` or `--pin-xip-sram` can be passed to add load map items indicating whether main SRAM should be cleared and whether XIP SRAM should be pinned.
```
:wink:

will-v-pi: I was not adding it here because we're not saying "all of SRAM", and there are plenty of other places in picotool & the SDK where "SRAM" means just "main SRAM"

lurch: Fair enough.

bintool/bintool.cpp
lurch: Does your "but may be more in future" comment mean that this code would be more "future-proof" if it iterated over `load_map->entries`, rather than just checking that `load_map->entries.size() == 1` and then looking only at `load_map->entries[0]`?
(I dunno much about load-maps, so this might be a stupid question :laughing: )

will-v-pi: No, that would catch load maps created by picotool as generic load maps - we can revisit if more generic load maps are added to the SDK, but for now this is correct for the XIP pinning load map

PR #291, opened 2026-02-11
Respect boolean values in partition table JSON

will-v-pi:
Previously picotool would treat presence of a key as true, and ignore the actual boolean value set

Fixes #290

PR #303, opened 2026-04-01
Check value of partition flags for ignoring boot

broughtong:
Spent a bit of time wondering why my partition boot flags were different to what I had specified!

In the JSON input file, we're currently simply checking if the flags `ignored_during_riscv_boot` and `ignored_during_arm_boot` are present. If so, they're ignored.

So this:

`"ignored_during_riscv_boot": true`

is currently evaluated exactly the same as this:

`"ignored_during_riscv_boot": false`

This PR adds a get_json_boolean function based on the get_json_int functions to parse the value part of the JSON key/value entry, and we now take the boolean value into account rather than just the if the key is present or not.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -1931,6 +1931,15 @@ bool get_json_int(json value, T& out) {
     }
 }
 
+bool get_json_boolean(json value, bool& out) {
+    if (value.is_boolean()) {
+        out = value.get<bool>();
+        return true;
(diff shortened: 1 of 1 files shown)


broughtong:
Seems it may have already been fixed in dev branch. Please ignore...

lurch:
Yup, fixed by #291

PR #306, opened 2026-04-23
Improve family ID CLI handling

### LINKED ISSUE #305: Improve discoverability of family names and improve the user experience in related error messages
I was just helping someone out with running `picotool uf2 convert` because it was complaining that the family ID `rp2350` was not valid. I've never used this tool before, so my first port of call was to run the help command. This did not provide any information on what the family ID should be for an RP2350. I looked at the README file on this repository, which did not appear to have a section listing the supported family names. I then tried searching the source code for `family_id`, `--family`, `rp2350`, and some of the constant values to try to discover what the flag should be. This did not help either - GitHub's search couldn't find where constants were defined, and I couldn't spot a location where all family names are enumerated. In the end I searched the closed issues for "rp2350" and found some examples that showed that the family ID should've been either `rp2350-riscv` or `rp2350-arm-s`. I think this user experience can be significantly improved.

The first thing I would like to request is a `picotool list families` command that lists the family IDs that are supported by the tool. This makes the families much more discoverable, directly within the tool. Implementing this as a `list` command (instead of, say, `picotool list-families`) provides a common basis for future functionality that lists other entities for easy discovery. Including this list in the readme would also be helpful, but this is secondary to having it directly in the tool.

The second thing I would like to request is a modification to the error message when an invalid family ID is provided. The error message should, of course, be updated to point the user to the new `picotool list families` command, instead of the more general help page. In addition, I think it would be helpful to perform a substring search on the erroneous family name that was provided by the user and return any matches as suggestions. For example, if a user enters `rp2350`, the error message should suggest "perhaps you meant one of these families: rp2350-arm-s, rp2350-riscv". Similarly, if a user enters `RP2350-ARM`, the error message should suggest "perhaps you meant rp2350-arm-s". This message would provide instant and helpful feedback without the user needing to execute a second command to discover the correct family name.

will-v-pi:
Add list of valid IDs to the error message when you enter an invalid ID

Handle uppercase and _ in the name (eg RP2350_ARM_S gets converted to rp2350-arm-s)

Also add this improved handling to partition table creation

Drive by fixes to prevent UF2 convert from touching files until input file type has been checked, and update json examples to point to master schemas instead of develop

Partially fixes #305

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -127,6 +127,36 @@ static const string rp2350_arm_ns_family_name = "rp2350-arm-ns";
 static const string rp2350_riscv_family_name = "rp2350-riscv";
 static const string cyw43_firmware_family_name = "cyw43-firmware";
 
+static const std::vector<std::string> family_names = {
+    data_family_name,
+    absolute_family_name,
+    rp2040_family_name,
+    rp2350_arm_s_family_name,
+    rp2350_arm_ns_family_name,
+    rp2350_riscv_family_name,
+    cyw43_firmware_family_name
+};
+
+static const std::map<uint32_t, std::string> family_id_to_name = {
+    {DATA_FAMILY_ID, data_family_name},
+    {ABSOLUTE_FAMILY_ID, absolute_family_name},
+    {RP2040_FAMILY_ID, rp2040_family_name},
+    {RP2350_ARM_S_FAMILY_ID, rp2350_arm_s_family_name},
+    {RP2350_ARM_NS_FAMILY_ID, rp2350_arm_ns_family_name},
+    {RP2350_RISCV_FAMILY_ID, rp2350_riscv_family_name},
+    {CYW43_FIRMWARE_FAMILY_ID, cyw43_firmware_family_name}
+};
+
+static const std::map<std::string, uint32_t> family_name_to_id = {
+    {data_family_name, DATA_FAMILY_ID},
+    {absolute_family_name, ABSOLUTE_FAMILY_ID},
+    {rp2040_family_name, RP2040_FAMILY_ID},
+    {rp2350_arm_s_family_name, RP2350_ARM_S_FAMILY_ID},
+    {rp2350_arm_ns_family_name, RP2350_ARM_NS_FAMILY_ID},
+    {rp2350_riscv_family_name, RP2350_RISCV_FAMILY_ID},
+    {cyw43_firmware_family_name, CYW43_FIRMWARE_FAMILY_ID}
+};
+
 #if !HAS_LIBUSB
 static const string built_without_libusb_message = "\nThis version of picotool was compiled without USB support. Some commands are not available.\n";
 #endif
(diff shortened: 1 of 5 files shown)


main.cpp
```diff
@@ -127,6 +127,36 @@ static const string rp2350_arm_ns_family_name = "rp2350-arm-ns";
...
+    {DATA_FAMILY_ID, data_family_name},
+    {ABSOLUTE_FAMILY_ID, absolute_family_name},
+    {RP2040_FAMILY_ID, rp2040_family_name},
+    {RP2350_ARM_S_FAMILY_ID, rp2350_arm_s_family_name},
+    {RP2350_ARM_NS_FAMILY_ID, rp2350_arm_ns_family_name},
+    {RP2350_RISCV_FAMILY_ID, rp2350_riscv_family_name},
+    {CYW43_FIRMWARE_FAMILY_ID, cyw43_firmware_family_name}
+};
+
+static const std::map<std::string, uint32_t> family_name_to_id = {
```
lurch, line 150: As `family_name_to_id` is simply the "inverse" of `family_id_to_name`, would it be worth building it programmatically? (instead of having to keep two separate mappings in sync)

main.cpp
```diff
@@ -365,21 +395,12 @@ struct family_id : public cli::value_base<family_id> {
...
-            } else if (value == rp2350_arm_s_family_name) {
-                t = RP2350_ARM_S_FAMILY_ID;
-            } else if (value == rp2350_arm_ns_family_name) {
-                t = RP2350_ARM_NS_FAMILY_ID;
-            } else if (value == rp2350_riscv_family_name) {
-                t = RP2350_RISCV_FAMILY_ID;
-            } else if (value == cyw43_firmware_family_name) {
-                t = CYW43_FIRMWARE_FAMILY_ID;
+            std::transform(value.begin(), value.end(), value.begin(),
+                [](unsigned char c){ return std::tolower(c); });
```
lurch, line 399: I'm not much of a C++ programmer, but is this really the easiest way of converting a string to lowercase? (Maybe I've become too used to Python :wink: )

main.cpp
```diff
@@ -365,21 +395,12 @@ struct family_id : public cli::value_base<family_id> {
...
+            std::replace( value.begin(), value.end(), '_', '-');
+            auto family_id = family_name_to_id.find(value);
+            if (family_id != family_name_to_id.end()) {
+                t = family_id->second;
             } else {
                 if (value.find("0x") == 0) {
```
lurch: Now that the previous if-stack has collapsed, perhaps it makes sense to convert this to an `else if`? :shrug:

main.cpp
```diff
@@ -6456,21 +6475,32 @@ uint32_t permissions_to_flags(json permissions) {
...
             ret |= PICOBIN_PARTITION_FLAGS_ACCEPTS_DEFAULT_FAMILY_RP2040_BITS;
-        } else if (family == rp2350_arm_s_family_name) {
+        } else if (id == RP2350_ARM_S_FAMILY_ID) {
             ret |= PICOBIN_PARTITION_FLAGS_ACCEPTS_DEFAULT_FAMILY_RP2350_ARM_S_BITS;
-        } else if (family == rp2350_arm_ns_family_name) {
+        } else if (id == RP2350_ARM_NS_FAMILY_ID) {
             ret |= PICOBIN_PARTITION_FLAGS_ACCEPTS_DEFAULT_FAMILY_RP2350_ARM_NS_BITS;
-        } else if (family == rp2350_riscv_family_name) {
+        } else if (id == RP2350_RISCV_FAMILY_ID) {
             ret |= PICOBIN_PARTITION_FLAGS_ACCEPTS_DEFAULT_FAMILY_RP2350_RISCV_BITS;
```
lurch, line 6499: Maybe it's worth making this a `std::map` too?

will-v-pi: I thought about that, but given it's only used for creating & parsing partition tables, I don't think it's worth it. It might also be confusing to maintain, because all the other maps are the same size, but this one would be smaller.

main.cpp
will-v-pi: I think it is, but feel free to add a suggestion if you find something simpler

lurch: Huh, so it is :joy:  https://www.geeksforgeeks.org/cpp/how-to-convert-std-string-to-lower-case-in-cpp/

main.cpp
lurch: Okay.

main.cpp
will-v-pi: The alternative would be keeping two separate lists (of names and family IDs) in sync, and for those the order would matter, so I think this is fine - also, building it at runtime might be slower to run?

main.cpp
will-v-pi: Done

main.cpp
lurch: Nah, I wasn't suggesting having two separate lists, and then building _both_ `family_name_to_id` and `family_id_to_name` programmatically, I was only suggesting building `family_name_to_id` from `family_id_to_name` (in the spirit of [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)).
But for such small maps it's probably not worth it, and you'd also lose the const-ness.

main.cpp
```diff
@@ -401,8 +401,7 @@ struct family_id : public cli::value_base<family_id> {
             auto family_id = family_name_to_id.find(value);
             if (family_id != family_name_to_id.end()) {
                 t = family_id->second;
-            } else {
-                if (value.find("0x") == 0) {
+            } else if (value.find("0x") == 0) {
                     value = value.substr(2);
                     size_t pos = 0;
```
lurch: I guess these lines ought to be de-dented?

lurch:
I was curious about the build errors, but upon investigation it looks like these will be fixed by https://github.com/raspberrypi/pico-sdk/pull/2906 :slightly_smiling_face:

PR #307, opened 2026-04-29
Remove chocolatey from Windows CI

will-v-pi:
`choco install` started failing, and GCC is already installed on Windows GitHub runners, so just use that


lurch:
LGTM as the tests are still passing :+1:

kilograham (approved):

will-v-pi:
Also using this PR to update Windows visual studio version, as 17 is no longer available on GitHub runners

PR #310, opened 2026-05-04
Fix typo in message about devices found

barnas-michal:
Example of message changed:
```
No accessible RP-series devices in BOOTSEL mode were found found with serial number XXX.
```

Duplicated "found" word.


lurch (approved):
Nice catch!

barnas-michal:
Thanks :)

PR #311, opened 2026-05-05
Don't fail CI when curl is unable to check CRL for the sample UF2 files

lurch:
The reason for this failure is unclear, but it's definitely not a fatal error

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -85,11 +85,11 @@ jobs:
       - name: Test
         run: |
           picotool help
-          curl -L https://datasheets.raspberrypi.com/soft/blink.uf2 -o blink.uf2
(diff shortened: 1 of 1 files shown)


will-v-pi:
I think the GitHub CI is just generally being a bit flaky at the moment, which could explain this - are there any jobs that haven't succeeded after a retry?

lurch:
> I think the GitHub CI is just generally being a bit flaky at the moment, which could explain this - are there any jobs that haven't succeeded after a retry?

Some of the failed jobs in #310 worked after a retry; but some of the failed jobs still failed after a retry.

lurch:
All the CI checks on #312 have passed, so it looks like this was just a temporary GitHub hiccup which has now been resolved.

PR #312, opened 2026-05-08
Prevent coprodis warning for .word

will-v-pi:
Prevent coprodis warning for .word that happen to be coprocessor instructions

Didn't affect GCC as words are formatted as 8 hex chars instead of 4+4, but did affect clang as they're formatted the same as instructions


main.cpp
```diff
@@ -7461,6 +7461,16 @@ bool coprodis_command::execute(device_map &devices) {
             }
             return pos - pos0;
         };
+        auto check_for_word = [&](size_t pos) {
```
kilograham: um this may read off the end of the line which isn't mull terminated... i would use string::compare ;-)

kilograham (changes requested):

main.cpp
will-v-pi: I've switched it over to use string::compare instead

kilograham (approved):

PR #313, opened 2026-05-14
Add uf2 combine command

### LINKED ISSUE #169: picotool uf2 convert cannot append arbitrary data
I’ve got 2 data files :
table.bin (283 bytes)
configuration.bin (64 bytes)

if I do
picotool uf2 convert -o 0x10001000 table.bin pico.uf2
picotool uf2 convert -o 0x10002000 configuration.bin pico.uf2
then I got :
a file pico.uf2 of 512 bytes long. It means table.bin has been removed from pico.uf2 (or pico.uf2 has been deleted before adding configuration.bin)

Thanks to picotool uf2 convert we should be able to add arbitrary data at arbitrary location.

I do not have this issue when I append binary codes compiled for the right location.

will-v-pi:
This combines 2 UF2s into 1, modifying the block numbers, and giving them same family ID, so it can be loaded as a single UF2 file. Can be called repeatedly to combine more UF2s.

Main use case is for combining partition tables with their partitions for loading as a single UF2 file. This will make running pico-examples that use partition tables easier, as a single UF2 can be generated for loading.

I think this was requested by @timg236, and it also fixes #169

### DIFF
--- a/README.md
+++ b/README.md
@@ -894,11 +894,15 @@ OPTIONS:
         <offset>
             Load offset (memory address; default 0x10000000 for BIN file)
     UF2 Family options
+        --family
+            Specify the family ID
         <family_id>
             family ID for UF2
     Platform options
+        --platform
+            Optional platform for memory verification
         <platform>
-            optional platform for memory verification (eg rp2040, rp2350)
+            platform to use (eg rp2040, rp2350)
     Errata RP2350-E10 Fix
         --abs-block
             Add an absolute block
@@ -940,6 +944,8 @@ OPTIONS:
         -t <type>
             Specify file type (uf2) explicitly, ignoring file extension
     UF2 Family options
+        --family
+            Specify the family ID
         <family_id>
             family ID for combined UF2 (defaults to first one)
     Offset options
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ SYNOPSIS:
     picotool reboot [-a] [-u] [-g <partition>] [-c <cpu>] [device-selection]
     picotool otp list|get|set|load|dump|permissions|white-label
     picotool partition info|create
-    picotool uf2 info|convert
+    picotool uf2 info|convert|combine
     picotool version [-s] [<version>]
     picotool coprodis [--quiet] [--verbose] <infile> <outfile>
     picotool help [<cmd>]
@@ -904,6 +904,101 @@ OPTIONS:
             Add an absolute block
         <abs_block_loc>
             absolute block location (default to 0x10ffff00)
+```
+
+### combine
+
+This command is used to combine multiple UF2 files (possibly with different family IDs) into a single file with the same family ID. This can be useful for loading a partition table and multiple partitions onto a device using a single file.
+
+```text
+$ picotool help uf2 combine
+UF2 COMBINE:
+    Combine multiple UF2 files.
+
+SYNOPSIS:
+    picotool uf2 combine [--quiet] [--verbose] <infile1> [-t <type>] <infile2> [-t <type>] <outfile> [-t <type>] [--family <family_id>]
+                [--offset <offset>] [--partition <partition>] [[--abs-block] [<abs_block_loc>]]
+
+OPTIONS:
+        --quiet
+            Don't print any output
+        --verbose
+            Print verbose output
+    First file to combine
+        <infile1>
+            The file name
+        -t <type>
+            Specify file type (uf2) explicitly, ignoring file extension
+    Second file to combine
+        <infile2>
+            The file name
+        -t <type>
+            Specify file type (uf2) explicitly, ignoring file extension
+    File to save output to
+        <outfile>
+            The file name
+        -t <type>
+            Specify file type (uf2) explicitly, ignoring file extension
+    UF2 Family options
+        <family_id>
+            family ID for combined UF2 (defaults to first one)
+    Offset options
+        --offset
+            Offset second UF2 by amount
+        <offset>
+            offset amount (default to 0)
+    Partition options
+        --partition
+            Place second UF2 in partition (first UF2 must contain a partition table)
+        <partition>
+            partition number (default to 0)
+    Errata RP2350-E10 Fix
+        --abs-block
+            Add an absolute block
+        <abs_block_loc>
+            absolute block location (default to 0x10ffff00)
+```
+
+The `--partition` argument can be used to place the second file in a partition number, provided that there is a partition table in the first file. For example, take this `pt.json`
+```json
+{
+    "unpartitioned": {
+        "families": ["absolute"],
+        "permissions": {
+            "secure": "rw",
+            "nonsecure": "rw",
+            "bootloader": "rw"
+        }
+    },
+    "partitions": [
+        {
+            "size": "1024K",
+            "families": ["rp2350-arm-s", "rp2350-riscv"],
+            "permissions": {
+                "secure": "rw",
+                "nonsecure": "rw",
+                "bootloader": "rw"
+            }
+        },
+        {
+            "size": "1024K",
+            "families": ["data"],
+            "permissions": {
+                "secure": "rw",
+                "nonsecure": "rw",
+                "bootloader": "rw"
+            }
+        }
+    ]
+}
+```
+
+If you want to load this partition table, and put `code.uf2` and `data.uf2` into the partitions, all using a single UF2, you would run:
+```text
+$ picotool partition create pt.json pt.uf2
+$ picotool uf2 combine pt.uf2 code.uf2 tmp.uf2 --partition 0
+$ picotool uf2 combine tmp.uf2 data.uf2 combined.uf2 --partition 1
+$ picotool load -x combined.uf2
 ```
 
 ### info
--- a/README.md
+++ b/README.md
@@ -962,7 +962,7 @@ OPTIONS:
             family ID for UF2
     Platform options
         --platform
-            Optional platform for memory verification
+            Optional platform for memory-address validation
         <platform>
             platform to use (eg rp2040, rp2350)
     Errata RP2350-E10 Fix
--- a/main.cpp
+++ b/main.cpp
@@ -1355,10 +1355,12 @@ struct uf2_convert_command : public cmd {
                         hex("offset").set(settings.offset) % "Load offset (memory address; default 0x10000000 for BIN file)"
                 ).force_expand_help(true) % "Packaging Options" + 
                 (
-                    option("--family") & family_id("family_id").set(settings.family_id) % "family ID for UF2"
+                    option("--family") % "Specify the family ID" &
+                        family_id("family_id").set(settings.family_id) % "family ID for UF2"
                 ).force_expand_help(true) % "UF2 Family options" +
                 (
-                    option("--platform") & platform_model("platform").set(settings.model) % "optional platform for memory verification (eg rp2040, rp2350)"
+                    option("--platform") % "Optional platform for memory verification" &
+                        platform_model("platform").set(settings.model) % "platform to use (eg rp2040, rp2350)"
                 ).force_expand_help(true) % "Platform options"
             #if SUPPORT_RP2350_A2
                 + (
@@ -1387,7 +1389,8 @@ struct uf2_combine_command : public cmd {
                 named_typed_file_selection_x("infile2", 2, "uf2") % "Second file to combine" +
                 named_typed_file_selection_x("outfile", 0, "uf2") % "File to save output to" +
                 (
-                    option("--family") & family_id("family_id").set(settings.family_id) % "family ID for combined UF2 (defaults to first one)"
+                    option("--family") % "Specify the family ID" &
+                        family_id("family_id").set(settings.family_id) % "family ID for combined UF2 (defaults to first one)"
                 ).force_expand_help(true) % "UF2 Family options" +
                 (
                     option("--offset").set(settings.uf2.offset_set) % "Offset second UF2 by amount" &
--- a/main.cpp
+++ b/main.cpp
@@ -1370,11 +1370,41 @@ struct uf2_convert_command : public cmd {
     }
 };
 
+struct uf2_combine_command : public cmd {
+    uf2_combine_command() : cmd("combine") {}
+    bool execute(device_map &devices) override;
+    virtual device_support get_device_support() override { return none; }
+
+    group get_cli() override {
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -906,6 +906,101 @@ OPTIONS:
...
+            The file name
+        -t <type>
+            Specify file type (uf2) explicitly, ignoring file extension
+    File to save output to
+        <outfile>
+            The file name
+        -t <type>
+            Specify file type (uf2) explicitly, ignoring file extension
+    UF2 Family options
+        <family_id>
```
lurch, line 1011: Should this be `--family <family_id>` ?

README.md
```diff
@@ -906,6 +906,101 @@ OPTIONS:
...
+    Partition options
+        --partition
+            Place second UF2 in partition (first UF2 must contain a partition table)
+        <partition>
+            partition number (default to 0)
+    Errata RP2350-E10 Fix
+        --abs-block
+            Add an absolute block
+        <abs_block_loc>
+            absolute block location (default to 0x10ffff00)
```
lurch, line 1027: Does this assume a particular-sized Flash is being used?

README.md
```diff
@@ -906,6 +906,101 @@ OPTIONS:
...
+    ]
+}
+```
+
+If you want to load this partition table, and put `code.uf2` and `data.uf2` into the partitions, all using a single UF2, you would run:
+```text
+$ picotool partition create pt.json pt.uf2
+$ picotool uf2 combine pt.uf2 code.uf2 tmp.uf2 --partition 0
+$ picotool uf2 combine tmp.uf2 data.uf2 combined.uf2 --partition 1
+$ picotool load -x combined.uf2
```
lurch, line 1069:
```suggestion
$ picotool uf2 combine tmp.uf2 data.uf2 combined.uf2 --partition 1
$ rm tmp.uf2
$ picotool load -x combined.uf2
```
:wink:

README.md
```diff
@@ -906,6 +906,101 @@ OPTIONS:
...
+        }
+    ]
+}
+```
+
+If you want to load this partition table, and put `code.uf2` and `data.uf2` into the partitions, all using a single UF2, you would run:
+```text
+$ picotool partition create pt.json pt.uf2
+$ picotool uf2 combine pt.uf2 code.uf2 tmp.uf2 --partition 0
+$ picotool uf2 combine tmp.uf2 data.uf2 combined.uf2 --partition 1
```
lurch, line 1068: Is there some theoretical future version of `picotool` where these separate steps might be able to be combined into a single command, or is it likely that these will always be separate steps? (Please forgive the naive question.)

README.md
will-v-pi: Will fix that, and the `uf2 convert` command I copied this from

README.md
will-v-pi: No - smaller flash mirrors at higher memory addresses (it's how picotool detects flash size), so this block always goes to the last block of flash regardless of the size

README.md
will-v-pi: `picotool` is cross platform, and `rm` isn't available on Windows CMD, so I think it's better to skip it

README.md
will-v-pi: Yes, but I think that requires CLI changes to support passing lists of files, so is outside the scope of this PR - `picotool link` is another command that would benefit from this

README.md
```diff
@@ -31,7 +31,7 @@ SYNOPSIS:
     picotool reboot [-a] [-u] [-g <partition>] [-c <cpu>] [device-selection]
     picotool otp list|get|set|load|dump|permissions|white-label
     picotool partition info|create
-    picotool uf2 info|convert
+    picotool uf2 info|convert|combine
```
lurch, line 34: Only a minor thing, but here it says `uf2 info|convert|combine`, but further down the page they're documented in the order: `convert`, `combine`, `info`. Could / should the sub-commands be listed in the same order in both places?

lurch:
> I think this was requested by @timg236, and it also fixes #169

I think @peterharperuk also said that it'd be nice if some of the pico-examples were also installable with a single UF2 file?

README.md
lurch: Yes, definitely outside the scope of this PR - that's why I said "theoretical future version" :wink:

README.md
will-v-pi: It shouldn't be too hard to adjust the order in the help text, as it's just changing the order in the vectors (I'd prefer that to changing the order in the README, as the README order is more logical) - but that can come in a separate PR, as lots more commands will need re-ordering, could you make an issue so I don't forget?

lurch: #315

lurch:
Could / should this PR also contain code to do https://github.com/raspberrypi/pico-examples/pull/746#issuecomment-4451310555 in order to remove the Unix-only limitation? Or is that outside the scope of what this PR is doing?

main.cpp
```diff
@@ -1351,10 +1355,12 @@ struct uf2_convert_command : public cmd {
                         hex("offset").set(settings.offset) % "Load offset (memory address; default 0x10000000 for BIN file)"
                 ).force_expand_help(true) % "Packaging Options" + 
                 (
-                    option("--family") & family_id("family_id").set(settings.family_id) % "family ID for UF2"
+                    option("--family") % "Specify the family ID" &
+                        family_id("family_id").set(settings.family_id) % "family ID for UF2"
                 ).force_expand_help(true) % "UF2 Family options" +
                 (
-                    option("--platform") & platform_model("platform").set(settings.model) % "optional platform for memory verification (eg rp2040, rp2350)"
+                    option("--platform") % "Optional platform for memory verification" &
```
lurch: What does "memory verification" mean? Is it actually verifying the memory of the RP2350, or is it just checking that addresses are in the right ranges? If the latter, perhaps "Optional platform for memory-address validation" would be better? :shrug:

main.cpp
```diff
@@ -3239,6 +3284,25 @@ std::unique_ptr<block> find_last_block(memory_access &raw_access, vector<uint8_t
     return nullptr;
 }
 
+std::vector<std::unique_ptr<block>> find_all_blocks(memory_access &raw_access, vector<uint8_t> &bin) {
+    // todo read the right amount
```
lurch, line 3327: Is this TODO a blocker for this PR being merged, or just a "something that it might be nice to fix in future"?

main.cpp
```diff
@@ -6775,6 +6839,120 @@ bool uf2_convert_command::execute(device_map &devices) {
     return false;
 }
 
+bool uf2_combine_command::execute(device_map &devices) {
+    if (get_file_type_idx(0) != filetype::uf2 || get_file_type_idx(1) != filetype::uf2 || get_file_type_idx(0) != filetype::uf2) {
```
```suggestion
    if (get_file_type_idx(0) != filetype::uf2 || get_file_type_idx(1) != filetype::uf2) {
```
Or do you just want to be **extra** sure that the first file is a UF2? :laughing:

main.cpp
will-v-pi: It's copied from `find_best_block` and `find_last_block` - not a blocker, just something that would be nice to fix in the future

main.cpp
```diff
@@ -6775,6 +6839,120 @@ bool uf2_convert_command::execute(device_map &devices) {
...
+            settings.uf2.abs_block = true;
+            settings.uf2.abs_block_loc = block.target_addr;
+            file->read((char*)&block, sizeof(uf2_block));
+        }
+    #endif
+
+        num_blocks += block.num_blocks;
+
+        if (!settings.family_id) {
+            settings.family_id = block.file_size;
```
lurch, line 6959: I've admittedly got no idea what this code is doing, but this line looks kinda odd? :thinking:  (Seems like `family_id` and `file_size` would be two _very_ different things?)

will-v-pi:
> Could / should this PR also contain code to do [raspberrypi/pico-examples#746 (comment)](https://github.com/raspberrypi/pico-examples/pull/746#issuecomment-4451310555) in order to remove the Unix-only limitation? Or is that outside the scope of what this PR is doing?

This is similar but separate, as they are different methods of combining UF2 files - that would be a separate `uf2 concat` command or similar, which would be outside the scope of this PR

main.cpp
will-v-pi: Naming is weird, but it is correct - see the [UF2 specification](https://github.com/microsoft/uf2#file-format)

main.cpp
will-v-pi: It's just checking that addresses are in the right ranges
```suggestion
                    option("--platform") % "Optional platform for memory-address validation" &
```

main.cpp
lurch: I see that the page you linked to says:

> File size or board family ID or zero

Thanks :slightly_smiling_face:

lurch:
> > Could / should this PR also contain code to do [raspberrypi/pico-examples#746 (comment)](https://github.com/raspberrypi/pico-examples/pull/746#issuecomment-4451310555) in order to remove the Unix-only limitation? Or is that outside the scope of what this PR is doing?
> 
> This is similar but separate, as they are different methods of combining UF2 files - that would be a separate `uf2 concat` command or similar, which would be outside the scope of this PR

Hmm, so from my limited understanding of picotool and UF2 files, would the idea then be that `uf2 combine` is used to combine e.g. two RP2350 UF2s into a single larger UF2 (i.e. each block from the input UF2s gets written to a different flash address), but `uf2 concat` would be used to combine a RP2350 UF2 and a RP2040 UF2 into a single larger UF2 that could then be written to either board? (i.e. each block in the input UF2s _might_ be targetting the same flash address)

will-v-pi:
> > > Could / should this PR also contain code to do [raspberrypi/pico-examples#746 (comment)](https://github.com/raspberrypi/pico-examples/pull/746#issuecomment-4451310555) in order to remove the Unix-only limitation? Or is that outside the scope of what this PR is doing?
> > 
> > 
> > This is similar but separate, as they are different methods of combining UF2 files - that would be a separate `uf2 concat` command or similar, which would be outside the scope of this PR
> 
> Hmm, so from my limited understanding of picotool and UF2 files, would the idea then be that `uf2 combine` is used to combine e.g. two RP2350 UF2s into a single larger UF2 (i.e. each block from the input UF2s gets written to a different flash address), but `uf2 concat` would be used two combine a RP2350 UF2 and a RP2040 UF2 into a single larger UF2 that could then be written to either board? (i.e. each block in the input UF2s _might_ be targetting the same flash address)

Yes, that is generally correct:
- `uf2 combine` is to combine UF2s into a single larger UF2 with **one family ID**
- `uf2 concat` would be to concatenate UF2s into a single larger UF2 with **multiple family IDs** (could be RP2040 + RP2350, but could also be RP2350 code + data to go into different partitions)

`uf2 concat` is possible already with standard tools (eg `cat`) so picotool support isn't really necessary, whereas `uf2 combine` requires reading & editing the file data (UF2 block headers, partition table parsing) so makes sense to be implemented in picotool.

lurch:
> > Hmm, so from my limited understanding of picotool and UF2 files, would the idea then be that `uf2 combine` is used to combine e.g. two RP2350 UF2s into a single larger UF2 (i.e. each block from the input UF2s gets written to a different flash address), but `uf2 concat` would be used two combine a RP2350 UF2 and a RP2040 UF2 into a single larger UF2 that could then be written to either board? (i.e. each block in the input UF2s _might_ be targetting the same flash address)
> 
> Yes, that is generally correct:
> 
>   * `uf2 combine` is to combine UF2s into a single larger UF2 with **one family ID**
> 
>   * `uf2 concat` would be to concatenate UF2s into a single larger UF2 with **multiple family IDs** (could be RP2040 + RP2350, but could also be RP2350 code + data to go into different partitions)
> 
> 
> `uf2 concat` is possible already with standard tools (eg `cat`) so picotool support isn't really necessary, whereas `uf2 combine` requires reading & editing the file data (UF2 block headers, partition table parsing) so makes sense to be implemented in picotool.

Now I'm slightly confused again - in the previous comment you said `uf2 concat` "could also be RP2350 code + data to go into different partitions", but in the opening comment for this PR (which adds `uf2 combine`) you said "Main use case is for combining partition tables with their partitions for loading as a single UF2 file" :confused:

will-v-pi:
> > > Hmm, so from my limited understanding of picotool and UF2 files, would the idea then be that `uf2 combine` is used to combine e.g. two RP2350 UF2s into a single larger UF2 (i.e. each block from the input UF2s gets written to a different flash address), but `uf2 concat` would be used two combine a RP2350 UF2 and a RP2040 UF2 into a single larger UF2 that could then be written to either board? (i.e. each block in the input UF2s _might_ be targetting the same flash address)
> > 
> > 
> > Yes, that is generally correct:
> > 
> > * `uf2 combine` is to combine UF2s into a single larger UF2 with **one family ID**
> > * `uf2 concat` would be to concatenate UF2s into a single larger UF2 with **multiple family IDs** (could be RP2040 + RP2350, but could also be RP2350 code + data to go into different partitions)
> > 
> > `uf2 concat` is possible already with standard tools (eg `cat`) so picotool support isn't really necessary, whereas `uf2 combine` requires reading & editing the file data (UF2 block headers, partition table parsing) so makes sense to be implemented in picotool.
> 
> Now I'm slightly confused again - in the previous comment you said `uf2 concat` "could also be RP2350 code + data to go into different partitions", but in the opening comment for this PR (which adds `uf2 combine`) you said "Main use case is for combining partition tables with their partitions for loading as a single UF2 file" 😕

They're both true - you can use:

1. `uf2 combine` to combine the partition table and the stuff in partitions into a single UF2 file with a single `absolute` family ID, which you can just load onto the device in one go
2. `uf2 concat` (literally just `cat` in this case, as you don't need the block-moving-hackery) to concatenate the UF2 files for the partitions into a single UF2 file with multiple family IDs, and provided you have setup the partition table correctly and it is **already loaded onto the device**, they will be loaded into their respective partitions in the order they appear in the file

1 is for writing everything, as it writes to absolute addresses, ignoring the partitions already on the device

2 is for updates, as it uses the bootrom to place the UF2s into the correct partitions as per their family IDs

Currently, you have to do 2, which involves loading the partition table onto the device, rebooting, then loading the concatenated UF2 onto the device.

This PR allows you to do 1, which loads the partition table and partition data with a single `picotool load`/drag & drop, ignoring what is already on the device

lurch:
Thanks William, I wasn't very familiar with how partitions / partition tables work on RP2350. But what you've written makes sense.

PR #314, opened 2026-05-15
Improve otp get

will-v-pi:
@peterharperuk raised the point that it currently prints the full 24-bit value for ECC rows, when it would be clearer to just print the 16-bit value - you can still output the full value by passing `--raw`


peterharperuk (approved):
Thanks - I think that makes things clearer

PR #316, opened 2026-05-29
Add support for bootsel activity LED

will-v-pi:
Can use `--led` and `--active-low` arguments with picotool commands that reboot the device to bootsel mode (ie `reboot -u` and forced commands), and can also set default LED with `DEFAULT_BOOTSEL_LED` to always flash that LED even when `--led` isn't passed

Also add a section to BUILDING.md about the various supported CMake variables that can be used to customise the build

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -106,8 +106,8 @@ The `picotool` build has some CMake variables you can use to customise how the b
 - `GENERATE_FIXED_DOCS_WIDTH`: By default the width of the output from `picotool` adjusts to the width of the terminal. Setting this to `true` fixes the width to 140, which is used when generating the [README](README.md).
 - `DEFAULT_BOOTSEL_LED`: This can be used to set the default value for the `--led` argument, so `picotool` always reboots devices to BOOTSEL with that LED flashing.
 - `PICOTOOL_NO_LIBUSB`: By default `picotool` is compiled with USB support if libusb is found. Setting this to `true` explicitly compiles without USB support, which is used when the Pico SDK builds picotool.
-- `USE_PRECOMPILED`: By default the build uses pre-compiles ELF/BIN files for code that is run on the device (enc_bootloader, xip_ram_perms, and picoboot_flash_id). Setting this to `false` re-compiles these files instead.
+- `USE_PRECOMPILED`: By default the build uses pre-compiled ELF/BIN files for code that is run on the device (enc_bootloader, xip_ram_perms, and picoboot_flash_id). Setting this to `false` re-compiles these files instead.
 
-These can all be set by passing `-DNAME=VALUE` to your `cmake` command (e.g. `-DGENERATE_FIXED_DOCS_WIDTH=1`).
+These can all be set by passing `-DNAME=VALUE` to your `cmake` command (e.g. `-DGENERATE_FIXED_DOCS_WIDTH=true`).
 
-Note that the behaviour when setting `PICOTOOL_NO_LIBUSB` or `USE_PRECOMPILED` is undefined when using an existing build directory, due to the use of CMake external projects - instead you should create a new build directory to ensure it is in the correct state.
+Note: when setting `PICOTOOL_NO_LIBUSB` or `USE_PRECOMPILED` you should create a new build directory (rather than using an existing build directory), due to the way `picotool` uses external CMake projects.
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -97,3 +97,17 @@ make install
 In order for the SDK to find `picotool` in this custom folder, you will usually need to set the `picotool_DIR` variable in your project. This can be achieved either by setting the `picotool_DIR` environment variable to `$MY_INSTALL_DIR/picotool`, by passing `-Dpicotool_DIR=$MY_INSTALL_DIR/picotool` to your `cmake` command, or by adding `set(picotool_DIR $MY_INSTALL_DIR/picotool)` to your CMakeLists.txt file.
 
 > See the [find_package documentation](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure) for more details
+
+
+## Customising the build
+
+The `picotool` build has some CMake variables you can use to customise how the build works, and how `picotool` functions:
+
+- `GENERATE_FIXED_DOCS_WIDTH`: By default the width of the output from `picotool` adjusts to the width of the terminal. Setting this to `true` fixes the width to 140, which is used when generating the [README](README.md).
+- `DEFAULT_BOOTSEL_LED`: This can be used to set the default value for the `--led` argument, so `picotool` always reboots devices to BOOTSEL with that LED flashing.
+- `PICOTOOL_NO_LIBUSB`: By default `picotool` is compiled with USB support if libusb is found. Setting this to `true` explicitly compiles without USB support, which is used when the Pico SDK builds picotool.
+- `USE_PRECOMPILED`: By default the build uses pre-compiles ELF/BIN files for code that is run on the device (enc_bootloader, xip_ram_perms, and picoboot_flash_id). Setting this to `false` re-compiles these files instead.
+
+These can all be set by passing `-DNAME=VALUE` to your `cmake` command (e.g. `-DGENERATE_FIXED_DOCS_WIDTH=1`).
+
+Note that the behaviour when setting `PICOTOOL_NO_LIBUSB` or `USE_PRECOMPILED` is undefined when using an existing build directory, due to the use of CMake external projects - instead you should create a new build directory to ensure it is in the correct state.
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -104,7 +104,7 @@ In order for the SDK to find `picotool` in this custom folder, you will usually 
 The `picotool` build has some CMake variables you can use to customise how the build works, and how `picotool` functions:
 
 - `GENERATE_FIXED_DOCS_WIDTH`: By default the width of the output from `picotool` adjusts to the width of the terminal. Setting this to `true` fixes the width to 140, which is used when generating the [README](README.md).
-- `DEFAULT_BOOTSEL_LED`: This can be used to set the default value for the `--led` argument, so `picotool` always reboots devices to BOOTSEL with that LED flashing.
+- `DEFAULT_BOOTSEL_LED`: This can be used to set the default value for the `--bootsel-led` argument, so `picotool` always reboots devices to BOOTSEL with that LED flashing.
 - `PICOTOOL_NO_LIBUSB`: By default `picotool` is compiled with USB support if libusb is found. Setting this to `true` explicitly compiles without USB support, which is used when the Pico SDK builds picotool.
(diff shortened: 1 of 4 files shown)


BUILDING.md
```diff
@@ -97,3 +97,17 @@ make install
...
+
+
+## Customising the build
+
+The `picotool` build has some CMake variables you can use to customise how the build works, and how `picotool` functions:
+
+- `GENERATE_FIXED_DOCS_WIDTH`: By default the width of the output from `picotool` adjusts to the width of the terminal. Setting this to `true` fixes the width to 140, which is used when generating the [README](README.md).
+- `DEFAULT_BOOTSEL_LED`: This can be used to set the default value for the `--led` argument, so `picotool` always reboots devices to BOOTSEL with that LED flashing.
+- `PICOTOOL_NO_LIBUSB`: By default `picotool` is compiled with USB support if libusb is found. Setting this to `true` explicitly compiles without USB support, which is used when the Pico SDK builds picotool.
+- `USE_PRECOMPILED`: By default the build uses pre-compiles ELF/BIN files for code that is run on the device (enc_bootloader, xip_ram_perms, and picoboot_flash_id). Setting this to `false` re-compiles these files instead.
```
lurch:
```suggestion
- `USE_PRECOMPILED`: By default the build uses pre-compiled ELF/BIN files for code that is run on the device (enc_bootloader, xip_ram_perms, and picoboot_flash_id). Setting this to `false` re-compiles these files instead.
```

BUILDING.md
```diff
@@ -97,3 +97,17 @@ make install
...
+
+These can all be set by passing `-DNAME=VALUE` to your `cmake` command (e.g. `-DGENERATE_FIXED_DOCS_WIDTH=1`).
```
lurch: Here you've got `-DGENERATE_FIXED_DOCS_WIDTH=1`, but a few lines earlier you said "`GENERATE_FIXED_DOCS_WIDTH`: ... Setting this to `true` ...".

If this is a case of "CMake treats 1 as true and 0 as false", then perhaps it makes sense to change the earlier phrasing to "Setting this to `1` ..." ?

BUILDING.md
```diff
@@ -97,3 +97,17 @@ make install
...
+
+Note that the behaviour when setting `PICOTOOL_NO_LIBUSB` or `USE_PRECOMPILED` is undefined when using an existing build directory, due to the use of CMake external projects - instead you should create a new build directory to ensure it is in the correct state.
```
lurch: Perhaps this would be clearer as:
```suggestion
Note: when setting `PICOTOOL_NO_LIBUSB` or `USE_PRECOMPILED` you should create a new build directory (rather than using an existing build directory), due to the way `picotool` uses external CMake projects.
```
rather than talking about "undefined behaviour"?

BUILDING.md
```diff
@@ -97,3 +97,17 @@ make install
...
 
 > See the [find_package documentation](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure) for more details
+
+
+## Customising the build
+
+The `picotool` build has some CMake variables you can use to customise how the build works, and how `picotool` functions:
+
+- `GENERATE_FIXED_DOCS_WIDTH`: By default the width of the output from `picotool` adjusts to the width of the terminal. Setting this to `true` fixes the width to 140, which is used when generating the [README](README.md).
+- `DEFAULT_BOOTSEL_LED`: This can be used to set the default value for the `--led` argument, so `picotool` always reboots devices to BOOTSEL with that LED flashing.
```
lurch: Should there also be a `DEFAULT_BOOTSEL_LED_ACTIVE_LOW` ? :thinking:

README.md
```diff
@@ -123,15 +123,18 @@ TARGET SELECTION:
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but without the
             USB drive mounted
+        --led <led>
+            Flash LED on this GPIO to show BOOTSEL activity (ignored by Arm RP2350-A2) - only applicable if this command reboots the device
```
lurch: Maybe?
```suggestion
            Flash LED on this GPIO to show BOOTSEL activity (ignored by RP2350-A2 in Arm mode) - only applicable if this command reboots the device
```

README.md
```diff
@@ -123,15 +123,18 @@ TARGET SELECTION:
...
+            to BOOTSEL mode
+        --active-low
+            The BOOTSEL activity LED is active low (ignored on RP2040 and RP2350-A4)
```
```suggestion
            The BOOTSEL activity LED is active low (ignored by RP2040 and RP2350-A4)
```
For consistency with the previous option :slightly_smiling_face:

README.md
```diff
@@ -123,15 +123,18 @@ TARGET SELECTION:
             Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the
             command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but without the
             USB drive mounted
+        --led <led>
```
lurch: Would it be possible to instead show
```suggestion
        --led <gpio>
```
?

Although thinking about it some more, perhaps the parameter names `--led` and `--active-low` are a little "too generic" (perhaps there might be other things which are "active low" in future?) and so `--bootsel-led` and `--bootsel-led-active-low` would be better names? :shrug:

lurch: Hmmm, does it make sense for the `picotool info` command to have a `--led` option??

main.cpp
```diff
@@ -633,6 +639,8 @@ auto device_selection =
         + option("--rp2040").set(settings.force_rp2040) % "Assume the device is an RP2040 - this is only required when using a custom vid/pid with an RP2040 on Windows, and is ignored on other operating systems"
         + option('f', "--force").set(settings.force) % "Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the command (unless the command itself is a 'reboot') the device will be rebooted back to application mode" +
                 option('F', "--force-no-reboot").set(settings.force_no_reboot) % "Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but without the USB drive mounted"
+        + (option("--led") & integer("led").set(settings.led)) % "Flash LED on this GPIO to show BOOTSEL activity (ignored by Arm RP2350-A2) - only applicable if this command reboots the device to BOOTSEL mode"
```
lurch: If `DEFAULT_BOOTSEL_LED` has been set (to a non-negative value), perhaps it would make sense to show that default value here?

main.cpp
```diff
@@ -8648,11 +8659,12 @@ bool reboot_command::execute(device_map &devices) {
         picoboot_memory_access raw_access(con);
         model_t model = raw_access.get_model();
         if (model->supports_picoboot_cmd(PC_REBOOT2)) {
+            uint32_t usb_flags = (settings.led >= 0) ? (settings.active_low ? 0x30u : 0x20u) : 0u;
```
lurch: Perhaps this could be
```suggestion
            uint32_t usb_flags = 0u;
            if (settings.led >= 0) {
                usb_flags |= BOOTSEL_FLAG_GPIO_PIN_SPECIFIED;
                if (settings.active_low)
                    usb_flags |= BOOTSEL_FLAG_GPIO_PIN_ACTIVE_LOW;
            }
```
to get rid of the magic numbers?

README.md
will-v-pi: Yes, because `picotool info -f` will reboot the device to bootsel mode if it isn't already, so will flash the LED when reading the info. These arguments are being added to all commands which interact with a device, as all of those have the possibility of rebooting it to bootsel mode.

BUILDING.md
will-v-pi: I think that's not necessary, especially because it would cause issues with turning it off (would need to invert the behaviour of the `--active-low` argument)

BUILDING.md
will-v-pi: CMake has lots of values it treats as `true` or `false` (https://cmake.org/cmake/help/latest/command/if.html#constant) - I'll change this example to `true` to match the text above

will-v-pi:
```suggestion
These can all be set by passing `-DNAME=VALUE` to your `cmake` command (e.g. `-DGENERATE_FIXED_DOCS_WIDTH=true`).
```

main.cpp
will-v-pi: Added, and re-worded the argument to make that simpler

kilograham (approved):

BUILDING.md
lurch: Good point, it hadn't occurred to me that `--active-low` just being a flag, rather than a parameter that takes an option, would make this "tricky" :slightly_smiling_face:

README.md
lurch: Cool, thanks for the clarification :+1:

BUILDING.md
```diff
@@ -97,3 +97,17 @@ make install
 In order for the SDK to find `picotool` in this custom folder, you will usually need to set the `picotool_DIR` variable in your project. This can be achieved either by setting the `picotool_DIR` environment variable to `$MY_INSTALL_DIR/picotool`, by passing `-Dpicotool_DIR=$MY_INSTALL_DIR/picotool` to your `cmake` command, or by adding `set(picotool_DIR $MY_INSTALL_DIR/picotool)` to your CMakeLists.txt file.
 
 > See the [find_package documentation](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure) for more details
+
+
+## Customising the build
+
+The `picotool` build has some CMake variables you can use to customise how the build works, and how `picotool` functions:
+
+- `GENERATE_FIXED_DOCS_WIDTH`: By default the width of the output from `picotool` adjusts to the width of the terminal. Setting this to `true` fixes the width to 140, which is used when generating the [README](README.md).
```
lurch: It's probably outside the scope of this PR, but I wonder if it might make sense to switch this from being a boolean to being an integer? So that you'd then do `-DGENERATE_FIXED_DOCS_WIDTH=140` rather than `-DGENERATE_FIXED_DOCS_WIDTH=true`?

BUILDING.md
```diff
@@ -97,3 +97,17 @@ make install
...
+- `DEFAULT_BOOTSEL_LED`: This can be used to set the default value for the `--bootsel-led` argument, so `picotool` always reboots devices to BOOTSEL with that LED flashing.
```
lurch: Perhaps "with that LED flashing" is a bit of a misnomer, and
```suggestion
- `DEFAULT_BOOTSEL_LED`: This can be used to set the default value for the `--bootsel-led` argument, so `picotool` always reboots devices to BOOTSEL with that LED used as the activity indicator.
```
would be better?

BUILDING.md
will-v-pi: I think keeping the default in the CMakeLists.txt file is better, so you don't have to look up the number to pass on the command line when you just want to update the readme. But maybe a way to override the default could be added using a different variable in another PR - although I'm not really sure what the use cases would be for that?

lurch: Ahh, if it's **only** used when updating the readme, then perhaps the existing behaviour doesn't need changing.

PR #320, opened 2026-06-01
Update pre-compiled ELFs using latest develop

will-v-pi:
Use new include-style linker scripts for enc_bootloader

`xip_ram_perms` can use new `xip_ram` binary type, which fixes `sleep_ms`

Also fixes a bug with `otp_permissions` where trying to set single digit pages with a leading zero (eg `09`) wouldn't work, and skip over any non-integer keys (eg `$schema`)


enc_bootloader/memmap_common/section_xip_ram.incl
```diff
@@ -0,0 +1 @@
+/* Don't use XIP_RAM */
```
lurch, line 1: Just out of curiosity, why `section_xip_ram.incl` (singular "section") but `sections_psram.incl` (plural "sections") ?

will-v-pi: Because the XIP_RAM one only contains one section, but the PSRAM one contains two - although maybe worth mentioning on the main PR, as the PSRAM one is an outlier

enc_bootloader/memmap_enc_bootloader/memory_ram.incl
```diff
@@ -0,0 +1,8 @@
+/* Entirely custom memory regions - scratch and ram together for ease */
+MEMORY
+{
+    RAM_START(rwx) : ORIGIN = 0x20080000, LENGTH = 0x44
+    SCRATCH_X(rwx) : ORIGIN = 0x20080044, LENGTH = 0xfbc
+    SCRATCH_Y(rwx) : ORIGIN = 0x20081000, LENGTH = 0x800
```
lurch, line 6: `SCRATCH_X` and `SCRATCH_Y` have different lengths??

will-v-pi: Yes - this is a heavily customised linker script, as the enc_bootloader requires memory in very specific places to avoid side channels

xip_ram_perms/CMakeLists.txt
```diff
@@ -20,8 +20,8 @@ if (NOT USE_PRECOMPILED)
     set(CMAKE_C_STANDARD 11)
     set(CMAKE_CXX_STANDARD 17)
 
-    if (PICO_SDK_VERSION_STRING VERSION_LESS "2.0.0")
-        message(FATAL_ERROR "Raspberry Pi Pico SDK version 2.0.0 (or later) required. Your version is ${PICO_SDK_VERSION_STRING}")
+    if (PICO_SDK_VERSION_STRING VERSION_LESS "2.2.1")
+        message(FATAL_ERROR "Raspberry Pi Pico SDK version 2.2.1 (or later) required. Your version is ${PICO_SDK_VERSION_STRING}")
```
lurch, line 24: Should something similar to this also be added to https://github.com/raspberrypi/pico-examples/pull/694 ? :thinking:

enc_bootloader/memmap_common/section_xip_ram.incl
lurch: Okay.

xip_ram_perms/CMakeLists.txt
will-v-pi: I think that generally gets updated separately when a new SDK is released, rather than for individual PRs - although this reminds me that I need to update enc_bootloader's CMakeLists.txt too

lurch: > I think that generally gets updated separately when a new SDK is released, rather than for individual PRs

To clarify, I meant specifically for `usb/device/dev_multi_cdc/CMakeLists.txt` rather than the top-level `CMakeLists.txt` (as I guess that'd still allow you build other parts of pico-examples with an older SDK, whilst ignoring  `dev_multi_cdc`? IYSWIM.)

will-v-pi: Generally these checks are project-wide - we wouldn't want to add an extra check for that specific example, as then it would need updating separately from the main one.

The reason these ELFs have separate checks it because they're CMake External Projects, and they're only compiled when `USE_PRECOMPILED=false`, so can be set to a higher version than the main `picotool` build.

lurch: Fair enough - as long as there's a good rationale for things being done the way they are, then I'm happy :slightly_smiling_face:

PR #322, opened 2026-06-09
Fix pins in binary info

### LINKED ISSUE #321: GPIO_FUNC_GPCK
If I add the following to my Pico 2 W program...

bi_decl(bi_1pin_with_func(RTC_CLOCK_SRC_GPIO_OUT, GPIO_FUNC_GPCK));

This get identifies as follows by picotool...

21:                     USB OVCUR DET

I suspect this is because of the differences between rp2040 and rp2350 where GPIO_FUNC_GPCK is 8 and 9 respectively. Can we account for the differences between chips here?

will-v-pi:
Propogate model to remapped_memory_access

Remove try catch, now model is returned as unknown

Use switch-case to select pins, and use empty pin functions for unknown rather than defaulting to RP2040

Fixes #321

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -175,6 +175,21 @@ static string hex_string(int64_t value, int width=8, bool prefix=true, bool uppe
     return ss.str();
 }
 
+std::array<std::array<string, 1>, 12> pin_functions_unknown{{
+    {""},
+    {""},
+    {""},
+    {""},
+    {""},
+    {""},
+    {""},
+    {""},
+    {""},
+    {""},
+    {""},
+    {""},
+}};
+
 std::array<std::array<string, 30>, 10> pin_functions_rp2040{{
     {"", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", ""},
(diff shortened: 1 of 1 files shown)


peterharperuk (approved):

main.cpp
```diff
@@ -2716,10 +2729,16 @@ struct bi_visitor_base {
...
+        switch (chip) {
+            case rp2040:
+                pins(pin_mask, func, name, pin_functions_rp2040);
+                break;
+            case rp2350:
+                pins(pin_mask, func, name, pin_functions_rp2350);
+                break;
+            default:
+                pins(pin_mask, func, name, pin_functions_unknown);
+                break;
```
lurch, line 2743: Could this just be
```suggestion
            default:
                break;
```
and then you could get rid of the (basically empty) `pin_functions_unknown` array?

will-v-pi: I added that array because I want it to print:
```
Fixed Pin Information
 0-1:    Unknown pin function 2
 3-4:    Unknown pin function 6
 32-33:  Unknown pin function 7
```

Otherwise it just prints:
```
Fixed Pin Information
 none
```

lurch: Ahhh! In that case is it important that `len(pin_functions_unknown) >= max(len(pin_functions_rp2040), len(pin_functions_rp2350))` ? (and if it is, _perhaps_ it's worth adding a `static_assert` ?)

will-v-pi: Added

lurch:
I've just given this a test, and it fixes the problem I was seeing in #321 :+1: 

```
$ picotool info -p pio_blink.uf2 
File pio_blink.uf2 family ID 'rp2350-arm-s':

Fixed Pin Information
 0:      UART0 TX
 1:      UART0 RX
 3-4:    PIO0
 32-33:  PIO1
```

lurch (approved):

lurch:
Should this be in the 2.2.1 milestone?

will-v-pi:
> Should this be in the 2.2.1 milestone?

Picotool doesn't have a 2.2.1 milestone

lurch:
Should it?

will-v-pi:
> Should it?

Maybe, I'm not sure - we've been less strict with milestones in `picotool` than `pico-sdk`, so there are lots of `picotool` PRs with no milestone. Note that the next SDK release will probably be 2.3.0 anyway as there have been a lot of changes, so the SDK milestones will get renamed

PR #325, opened 2026-06-16
Add ELF squashing

will-v-pi:
Adds squashing of segments in the ELF file together, for them to be copied to the correct place at runtime

This is necessary for signing/hashing of sparse no_flash binaries (eg using XIP_SRAM or SCRATCH_X/Y)

This behaviour is enabled by default, but a `--no-squash` argument is added which disables this squashing if necessary (eg if the pico-sdk binary is compiled with `PICO_CRT0_DEBUG_ENTRY_RESETS_VIA_BOOTROM=0` and `PICO_CRT0_NO_DATA_COPY=1`, as then the copy at runtime will not occur)


lurch:
Squashed Elf :joy: 
image

bintool/bintool.cpp
```diff
@@ -687,19 +676,51 @@ void hash_andor_sign_block(block *new_block, const public_t public_key, const pr
 }
 
 
+bool detect_generic_load_map_entry(const load_map_item::entry& entry, model_t model, bool &pin_xip_sram) {
```
lurch, line 679: Looks a bit weird to have a function that returns the same `bool` both via a reference and also via the return? :confused: (Perhaps it'd be better to just rely on the reference and have a `void` return-type?)

bintool/bintool.cpp
```diff
@@ -687,19 +676,51 @@ void hash_andor_sign_block(block *new_block, const public_t public_key, const pr
...
-    pin_xip_sram = load_map->entries.size() == 1
-        && load_map->entries[0].storage_address == 0x0
-        && load_map->entries[0].runtime_address == model->xip_sram_start()
-        && load_map->entries[0].size == 0x0;
+    if (load_map->entries.size() == 1) detect_generic_load_map_entry(load_map->entries[0], model, pin_xip_sram);
+
     return pin_xip_sram;
 }
 
+void remove_non_generic_load_map_entries(block *new_block, model_t model) {
```
lurch, line 701: Does this function specifically only work for "new" blocks, or does it work for any kind of `block`? (i.e. is the parameter-name misleading?)

elf/elf_file.cpp
```diff
@@ -533,6 +533,83 @@ const elf32_ph_entry* elf_file::segment_from_section(const elf32_sh_entry &sh) {
...
+
+    uint32_t last_seg_end = 0;
+
+    std::vector<elf32_ph_entry *> xip_sram_segs = {};
+
+    auto squash_seg = [&](elf32_ph_entry *seg) {
+        const uint32_t paddr = seg->physical_address();
+        const uint32_t psize = seg->physical_size();
+        if (!seg->is_load()) return;
+        if (!psize) return;
```
lurch: _Super_ minor style thing, but you've got `if (psize == 0)` on line 563 but `if (!psize)` here?

elf/elf_file.cpp
```diff
@@ -533,6 +533,83 @@ const elf32_ph_entry* elf_file::segment_from_section(const elf32_sh_entry &sh) {
...
+    if (highest_flash_address != 0) {
+        // cannot squash flash binaries
+        return;
+    }
+
+    uint32_t last_seg_end = 0;
+
+    std::vector<elf32_ph_entry *> xip_sram_segs = {};
+
+    auto squash_seg = [&](elf32_ph_entry *seg) {
```
lurch, line 580: It looks like this is just `move_seg` rather than `squash_seg` ?

main.cpp
```diff
@@ -6645,13 +6653,14 @@ bool partition_create_command::execute(device_map &devices) {
     elf_file *elf = &source_file;
     std::shared_ptr<block> pt_block;
     if (!settings.filenames[2].empty()) {
+        model_t model = get_model(2);
```
lurch, line 6656: What does this "magic number" 2 relate to? (Similarly, what does the 0 in `get_model(0)` on line 5885 relate to?)

README.md
```diff
@@ -26,8 +26,8 @@ SYNOPSIS:
     picotool erase -p <partition> [device-selection]
     picotool erase -r <from> <to> [device-selection]
     picotool reboot [-a] [-u] [-g <partition>] [-c <cpu>] [device-selection]
-    picotool seal [--quiet] [--verbose] [--hash] [--sign] [--clear] [--pin-xip-sram] <infile> [-t <type>] [-o <offset>] <outfile> [-t
-                <type>] <key> <otp> [--major <major>] [--minor <minor>] [--rollback <rollback> [<rows>..]]
+    picotool seal [--quiet] [--verbose] [--hash] [--sign] [--clear] [--pin-xip-sram] [--no-squash] <infile> [-t <type>] [-o <offset>]
```
lurch, line 29: This might be a stupid question, but why is it _only_ the `seal` command that has a `--no-squash` option?

main.cpp
will-v-pi: It’s the file_idx, so file number 2, matching the `filenames[2]` above - the other one is file number 0

README.md
will-v-pi: You only need to squash it once, before sealing it with a load map, hence doing the squashing in `seal` - none of the other commands squash the ELF, so don’t need the `no-squash` argument

bintool/bintool.cpp
will-v-pi: These functions all take the “new” block that is currently being created, hence using `new_block` everywhere

elf/elf_file.cpp
will-v-pi: It’s `move_seg_to_end_of_last_seg`, which is what I’m calling “squashing” in this PR

elf/elf_file.cpp
will-v-pi:
```suggestion
        if (psize == 0) return;
```

bintool/bintool.cpp
will-v-pi: It matches `detect_generic_load_map`, and see its use on line 714 for why this signature is useful - I’m keeping it this way in case we add more generic load maps in the future

lurch: > I’m keeping it this way in case we add more generic load maps in the future

Right, now I remember a similar conversation coming up before...

elf/elf_file.cpp
lurch: Haha, fair enough :grinning:

main.cpp
lurch: Ohh! I was worried that e.g. `get_model(0)` was hardcoded to return (the model for) RP2040 and `get_model(2)` was hardcoded to return RP2350B :wink: 
Probably outside the scope of this PR, but maybe `get_model_for_file_index(2)` would be less confusing?

README.md
lurch: Oh, so only a "sealed ELF" needs to be squashed? There wouldn't be any benefit to "squashing" a non-sealed ELF to e.g. make a smaller UF2 file? (apologies for my probably-horrendous use of terminology)

will-v-pi: That's correct - the UF2 file is already sparse, so squashing the ELF doesn't make it smaller.

This squashing is specifically useful for Packaged SRAM binaries which use XIP SRAM or Scratch X/Y (or other sparse SRAM layouts) - without squashing these will take up a large amount of space in Flash as the sparse bits of SRAM have to be stored with the correct gaps between them (ones using XIP SRAM won't even work, because XIP SRAM is before Main SRAM so the block loop won't be found), but with squashing they can be next to each other in Flash and then loaded into the correct bits of SRAM by the load map, taking up a lot less space.

main.cpp
will-v-pi: The argument is `file_idx`, but yes, could be clearer - feel free to make an issue to tidy it up. Note also `get_family_id`.

README.md
lurch: I don't fully understand all of that, but it sounds reasonable :grinning:

PR #326, opened 2026-06-18
Reduce default-pt partition sizes

will-v-pi:
Needs 3 * 4K at the end for BT flash bank and E10 abs block, in addition to the existing 2 * 4K at the start for PT slots

Also adds a warning to `picotool partition create` when this space isn't free, which will print like this - a bit wordy, happy to reduce the wordiness if there are suggestions
```
WARNING: This partition table covers the last 2 Flash sectors for flash size 4M, which will interfere with the BTStack flash bank if that is the actual flash size
```
or
```
WARNING: This partition table covers the last 3 Flash sectors for flash size 4M, which will interfere with the BTStack flash bank, and the RP2350-E10 absolute block, if that is the actual flash size
```

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -6621,6 +6621,17 @@ uint32_t families_to_flags(std::vector<string> families, bool fail_invalid = fal
     return ret;
 }
 
+uint32_t round_up_power_2(uint32_t v) {
+    v--;
+    v |= v >> 1;
+    v |= v >> 2;
+    v |= v >> 4;
+    v |= v >> 8;
+    v |= v >> 16;
+    v++;
+    return v;
+}
+
 bool partition_create_command::execute(device_map &devices) {
     if (get_file_type_idx(0) != filetype::json) {
         fail(ERROR_ARGS, "json must be a json file\n");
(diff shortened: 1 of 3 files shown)


lurch:
So I guess you'd get this warning if your UF2 file happened to come out at 4MB, but you've actually got 8MB of Flash on your RP2350 board? Is it worth adding a `--flash-size N` argument (so that picotool knows the real flash size) to avoid the warning in that case? (Probably not, but I thought I'd ask anyway :wink: )

will-v-pi:
> So I guess you'd get this warning if your UF2 file happened to come out at 4MB, but you've actually got 8MB of Flash on your RP2350 board? Is it worth adding a `--flash-size N` argument (so that picotool knows the real flash size) to avoid the warning in that case? (Probably not, but I thought I'd ask anyway 😉 )

I think it's fine - that is the reason I made this a WARNING rather than an ERROR, and the way to avoid it is just make your partition table a little bit bigger so it uses more than 4MB of Flash (or just ignore the warning as "if that is the actual flash size" is false)

main.cpp
kilograham: i guess this should say "Partiiton start"

main.cpp
```diff
@@ -6761,6 +6772,31 @@ bool partition_create_command::execute(device_map &devices) {
         pt.partitions.push_back(new_p);
     }
 
+    // Catch partition tables which cover the end of Flash, which the SDK uses for the BTStack flash bank
```
kilograham, line 6779: i'm a bit confused by this... cur_pos is just the end of the last partition, which needn't be in order, so presumablt this should be in the loop - not sure where the power of 2 comes in there - should we not be looking at FLASH_SIZE and comparing each partition

Additionally, it feels a bit ugly to have this, as we have no idea if the user has a btstack flash bank, but i can see the point. we should probably however add support to btstack_flash_bank to use a partition with a well known id (or family-id)

will-v-pi: I can change it to take the max value of `cur_pos` instead, which should deal with out-of-order partitions - don't want to compare each partition, because you might have a subsequent partition at a higher address (showing that your flash size is larger), so the comparison should happen at the end using the highest address used.

The power of 2 is to guess the minimum flash size that this partition table fits in (assuming that flash size is a power of 2, which is already assumed elsewhere in picotool), because `partition create` doesn't know what device it's being flashed to, so doen't know the actual flash size.

I agree about adding support for btstack_flash_bank to use a partition with a well known ID - when that's added, this check can be skipped if a partition with that ID is in the partition table.

lurch: I wonder if it might be worth having an option to say something like `--no-bluetooth`, and then this code in picotool would only need to check for the space for the E10 workaround? (But again, as this is only a warning and not an error, perhaps that'd just be unnecessary complication?)

will-v-pi: I think it's fine, as it's a warning not an error - I think it's better to inform the user of any possible issues with the PT, and they can always pass `--quiet` to silence the output if they don't care

kilograham: i wonder if we should add a specific option for quieting this particular warning

kilograham: (without quieting the others)

will-v-pi: Added `--no-btstack-flash-bank` to disable the flash bank part of the warning, and you can already pass `--abs-block 0` to disable the E10 abs block part of the warning

will-v-pi:
Confirmed that these warnings will not be printed when building as part of the SDK, as the SDK passes `--quiet` which silences all output from `picotool partition create` (other than errors)

PR #327, opened 2026-06-18
Add warning for boards missing bootsel resistor

will-v-pi:
This adds a warning to `picotool info` which gets printed when no partition table was loaded at boot, but a partition table is present in Flash:

```
WARNING: Device has no partition table loaded, but has a partition table at the start of flash.
         This may cause picotool to behave unpredictably.
         This could be due to using a custom board which is missing the resistor inline with the BOOTSEL button, or if
         BOOT_FLAGS0.HASHED_PARTITION_TABLE or BOOT_FLAGS0.SECURE_PARTITION_TABLE is set in OTP and the partition table is not hashed/signed
```

The cause for this that prompted this PR was a board missing the resistor inline with the BOOTSEL button, which meant CS stays low while the button is held, so the bootrom reads 0s from flash when booting up rather than reading the partition table. According to the hardware team, this can be a common issue with custom boards. Once the button is released, the partition table can then be read by `picotool info`.

This can also occur if you have an unhashed/unsigned PT with `BOOT_FLAGS0.HASHED_PARTITION_TABLE` or `BOOT_FLAGS0.SECURE_PARTITION_TABLE` set in OTP, hence adding those to the warning message.

Also fixes hash/sig verification for partition tables, and prints a warning when an incorrectly hashed/signed PT is found at the start of flash:

```
WARNING: Device has no partition table loaded, but has an incorrectly hashed/signed partition table at the start of flash.
         This may cause picotool to behave unpredictably.
```

Note that this warning does not occur when the partition table has been modified since load - that still throws the immediate error:
```
ERROR: The RP2350 device returned an error: modified data (pt modified since load)
```

The main unpredictable behaviour I'm referring to is the fact that `picotool partition info` says `there is no partition table` but `picotool info` prints the details of the partition table. `picotool` will also treat a device in this state as if it has no partition table, so will overwrite the partition table with `picotool load` rather than loading into a partition.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -3416,12 +3416,16 @@ static chip_t image_type_exe_chip_to_chip(uint image_type_exe_chip) {
 }
 
 #if HAS_LIBUSB
-void info_guts(memory_access &raw_access, picoboot::connection *con) {
+void info_guts(memory_access &raw_access, picoboot::connection *con, bool no_pt_loaded=false) {
+    model_t model = raw_access.get_model();
 #else
-void info_guts(memory_access &raw_access, void *con) {
+void info_guts(memory_access &raw_access, void *con, bool no_pt_loaded=false) {
 #endif
     // Use flash caching
     settings.use_flash_cache = true;
+    // Track if a valid partition table was found in the block loop
+    bool has_valid_pt = false;
+    bool has_invalid_pt = false;
     // Callback to pass to bintool, to get more bin data
     get_more_bin_cb more_cb = [&raw_access](std::vector<uint8_t> &bin, uint32_t offset, uint32_t size) {
         DEBUG_LOG("Now reading from %x size %x\n", offset, size);
@@ -3459,6 +3463,7 @@ void info_guts(memory_access &raw_access, void *con) {
             }
         };
         auto info_metadata = [&](block *current_block, bool verbose_metadata = false) {
+            bool is_pt = false;
             verified_t hash_verified = none;
             verified_t sig_verified = none;
         #if HAS_MBEDTLS
@@ -3514,6 +3519,8 @@ void info_guts(memory_access &raw_access, void *con) {
             // Partition Table
             auto partition_table = current_block->get_item<partition_table_item>();
             if (partition_table != nullptr) {
+                is_pt = true; // this block is a partition table
+                has_valid_pt = true; // assume it is valid - hash and sig checks will invalidate it if not
                 if (verbose_metadata) info_pair("block type", "partition table");
                 info_pair("partition table", partition_table->singleton ? "singleton" : "non-singleton");
                 std::stringstream unpartitioned;
@@ -3642,6 +3649,10 @@ void info_guts(memory_access &raw_access, void *con) {
                     }
                     info_pair("hash value", val.str());
                 }
+                if (is_pt && hash_verified != passed) {
+                    has_valid_pt = false; // partition table has bad hash
+                    has_invalid_pt = true;
+                }
             }
             if (sig_verified != none) {
                 info_pair("signature", sig_verified == passed ? "verified" : "incorrect");
@@ -3658,6 +3669,10 @@ void info_guts(memory_access &raw_access, void *con) {
                         pkey << hex_string(i, 2, false, true);
                     }
                     info_pair("public key", pkey.str());
+                }
+                if (is_pt && sig_verified != passed) {
+                    has_valid_pt = false; // partition table has bad signature
+                    has_invalid_pt = true;
                 }
             }
         };
@@ -3879,7 +3894,6 @@ void info_guts(memory_access &raw_access, void *con) {
         std::vector<std::pair<string,string>> device_state_pairs;
         if ((settings.info.show_device || settings.info.all) && raw_access.is_device()) {
             select_group(device_info);
-            model_t model = raw_access.get_model();
             uint8_t rom_version;
             raw_access.read_raw(0x13, rom_version);
             info_pair("type", model->name());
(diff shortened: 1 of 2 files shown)


lurch:
> Also fixes hash/sig verification for partition tables, and prints a warning when an incorrectly hashed/signed PT is found at the start of flash:
> 
> ```
> WARNING: Device has no partition table loaded, but has an incorrectly hashed/signed partition table at the start of flash.
>          This may cause picotool to behave unpredictably.
> ```

Will an incorrectly hashed/signed PT at the start of flash **always** lead to the PT not being loaded? If so, perhaps this message could be reduced to:
```
WARNING: Device has an incorrectly hashed/signed partition table at the start of flash.
         This may cause picotool to behave unpredictably.
```
?

will-v-pi:
> > Also fixes hash/sig verification for partition tables, and prints a warning when an incorrectly hashed/signed PT is found at the start of flash:
> > ```
> > WARNING: Device has no partition table loaded, but has an incorrectly hashed/signed partition table at the start of flash.
> >          This may cause picotool to behave unpredictably.
> > ```
> 
> Will an incorrectly hashed/signed PT at the start of flash **always** lead to the PT not being loaded? If so, perhaps this message could be reduced to:
> 
> ```
> WARNING: Device has an incorrectly hashed/signed partition table at the start of flash.
>          This may cause picotool to behave unpredictably.
> ```
> 
> ?

Yes, it should do - I will update the message

main.cpp
```diff
@@ -4063,6 +4077,24 @@ void info_guts(memory_access &raw_access, void *con) {
...
+        if (has_valid_pt) {
+            fos.first_column(0); fos.hanging_indent(0);
+            fos << "\nWARNING: Device has no partition table loaded, but has a partition table at the start of flash.\n";
+            fos.first_column(sizeof("WARNING: ") - 1);
+            fos << "This may cause picotool to behave unpredictably.\n";
+            fos << "This could be due to using a custom board which is missing the resistor inline with the BOOTSEL button, ";
+            fos << "or if BOOT_FLAGS0.HASHED_PARTITION_TABLE or BOOT_FLAGS0.SECURE_PARTITION_TABLE is set in OTP and the partition table is not hashed/signed\n";
+        } else if (has_invalid_pt) {
+            fos.first_column(0); fos.hanging_indent(0);
+            fos << "\nWARNING: Device has no partition table loaded, but has an incorrectly hashed/signed partition table at the start of flash.\n";
```
will-v-pi:
```suggestion
            fos << "\nWARNING: Device has an incorrectly hashed/signed partition table at the start of flash.\n";
```

main.cpp
```diff
@@ -3416,12 +3416,16 @@ static chip_t image_type_exe_chip_to_chip(uint image_type_exe_chip) {
...
+    model_t model = raw_access.get_model();
 #else
-void info_guts(memory_access &raw_access, void *con) {
+void info_guts(memory_access &raw_access, void *con, bool no_pt_loaded=false) {
 #endif
     // Use flash caching
     settings.use_flash_cache = true;
+    // Track if a valid partition table was found in the block loop
+    bool has_valid_pt = false;
+    bool has_invalid_pt = false;
```
lurch: Huh? One of these variables would seem to be redundant, given that a `bool` can only be `true` or `false`? :wink: 
Or if they're actually indicating slightly different things, then maybe one of them ought to be renamed for clarity?

will-v-pi: `has_valid_pt` is whether it has a valid PT

`has_invalid_pt` is whether it has an invalid PT

Neither means it has no PT

lurch: Ahhh. In that case maybe
```c
    bool has_pt = false;
    bool pt_is_valid = true;
```
would be more logical / less confusing-at-first-glance?

will-v-pi: I've renamed them

main.cpp
```diff
@@ -4080,14 +4078,14 @@ void info_guts(memory_access &raw_access, void *con, bool no_pt_loaded=false) {
 
 #if HAS_LIBUSB
     if (no_pt_loaded && model->chip() != rp2040) {
-        if (has_valid_pt) {
+        if (has_pt && pt_is_valid) {
```
lurch: Seems like this could be
```
if (has_pt) {
    if (pt_is_valid) {
        ...
    } else {
        ...
    }
}
```

will-v-pi: Yes, but I think I prefer the `has_pt && pt_is_valid` and `has_pt && (!pt_is_valid)` as it's clearer exactly what is being detected (as it's more similar to the variables I was using before) - it also reduces the level of indentation

lurch: Fair enough :slightly_smiling_face:

lurch: _Alternatively_, I guess the `if (no_pt_loaded && model->chip() != rp2040) {` line could be changed to `if (has_pt && no_pt_loaded && model->chip() != rp2040) {` and that way the indentation wouldn't change :wink:  (because it seems like `has_pt && no_pt_loaded` is kinda what this PR is all about?)

will-v-pi: That does sound clearer - added

main.cpp
```diff
@@ -4063,6 +4075,24 @@ void info_guts(memory_access &raw_access, void *con) {
...
+
+#if HAS_LIBUSB
+    if (has_pt && no_pt_loaded && model->chip() != rp2040) {
+        if (pt_is_valid) {
+            fos.first_column(0); fos.hanging_indent(0);
+            fos << "\nWARNING: Device has no partition table loaded, but has a partition table at the start of flash.\n";
+            fos.first_column(sizeof("WARNING: ") - 1);
+            fos << "This may cause picotool to behave unpredictably.\n";
+            fos << "This could be due to using a custom board which is missing the resistor in series with the BOOTSEL button, ";
+            fos << "or if BOOT_FLAGS0.HASHED_PARTITION_TABLE or BOOT_FLAGS0.SECURE_PARTITION_TABLE is set in OTP and the partition table is not hashed/signed\n";
```
lurch, line 4087: In the `else` case, is it worth mentioning which of these OTP flags are set? :shrug:

will-v-pi: The OTP flags won't affect the `else` case, that's to catch an incorrect hash/signature (e.g. a bit flip or something)

lurch: Okay - I guess if the partition table is invalid it's probably not actually helpful to know if it was the hash or the signature that was invalid?

PR #341, opened 2026-07-14
Better fix for elf squashing

will-v-pi:
Don't squash into occupied areas, and don't squash metadata blocks

This reverts commit 0d160860227c83133502b8e1c504313e40ae9124 which was an incomplete fix for this

### DIFF
--- a/bintool/bintool.cpp
+++ b/bintool/bintool.cpp
@@ -284,44 +284,11 @@ block place_new_block(elf_file *elf, std::unique_ptr<block> &first_block, model_
     } else {
         DEBUG_LOG("There is already a block loop\n");
         if (set_others_ignored) set_block_ignored(elf, first_block->physical_addr);
-        uint32_t next_block_addr = first_block->physical_addr + first_block->next_block_rel;
-        while (true) {
-            auto segment = elf->segment_from_physical_address(next_block_addr);
-            if (segment == nullptr) {
-                fail(ERROR_NOT_POSSIBLE, "The ELF file does not contain the next block address %x", next_block_addr);
-            }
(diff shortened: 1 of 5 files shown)


main.cpp
```diff
@@ -6177,8 +6177,20 @@ bool seal_command::execute(device_map &devices) {
         elf_file *elf = &source_file;
         elf->read_file(get_file(ios::in|ios::binary));
         if (!settings.seal.no_squash) {
+            // Segments containing a block that is already part of a metadata block loop must not be
+            // moved, as other blocks in the loop link to them by absolute physical address
```
lurch, line 6181: Only asking out of curiosity: would/could also rewriting the addresses in the other metadata blocks in the loop also be a solution here?

will-v-pi: Yes, but it's not necessary - this squashing is intended for large squashing distances (e.g. Scratch X/Y) or required squashing (e.g. XIP SRAM), not for squashing parts of a normal binary slightly closer together, so metadata blocks should not be moved in the intended squashing scenarios

bintool/bintool.cpp
```diff
@@ -362,6 +362,54 @@ block place_new_block(elf_file *elf, std::unique_ptr<block> &first_block, model_
 }
 
 
+std::vector<std::unique_ptr<block>> get_all_blocks(elf_file *elf, std::unique_ptr<block> &first_block) {
```
kilograham, line 335: probably worth doing what the bootrom does which is to detect cycles (can either use the. simple check the bootrom does which is to assume blocks are in order, or just use a set!)

will-v-pi: Added, using a set

PR #342, opened 2026-07-14
Add picotool help topics

### LINKED ISSUE #305: Improve discoverability of family names and improve the user experience in related error messages

will-v-pi:
The device selection options have got quite long, so move them into a separate `picotool help device-selection` command, rather than appearing in every help output

Also add a family-ids topic to list available family IDs (fixes raspberrypi/picotool#305)

Also makes various improvements to the help output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -981,6 +981,8 @@ SUB COMMANDS:
     permissions   Set the OTP access permissions. (RP2350 only)
     dump          Dump entire OTP. (RP2350 only)
     list          List matching known registers/fields.
+
+Use "picotool help otp <subcmd>" for more info
 ```
 
 ### get/set
@@ -1403,6 +1405,8 @@ SUB COMMANDS:
     rm       Delete a file or an empty directory on the block device.
     cat      Print contents of file on the block device.
     format   Format the block device.
+
+Use "picotool help bdev <subcmd>" for more info
 ```
 
 ### ls
--- a/README.md
+++ b/README.md
@@ -65,6 +65,10 @@ COMMANDS:
     link        Link multiple binaries into one block loop.
     bdev        Commands related to embedded block devices
 
+TOPICS:
+    device-selection   Options for Target Device Selection
+    family-ids         Valid Family IDs
+
 Use "picotool help <cmd>" for more info
 ```
 
@@ -117,34 +121,7 @@ OPTIONS:
 
 TARGET SELECTION:
     To target one or more connected RP-series device(s) in BOOTSEL mode (the default)
-        --bus <bus>
-            Filter devices by USB bus number
-        --address <addr>
-            Filter devices by USB device address
-        --vid <vid>
-            Filter by vendor id
-        --pid <pid>
-            Filter by product id
-        --ser <ser>
-            Filter by serial number
-        --rp2040
-            Assume the device is an RP2040 - this is only required when using a custom vid/pid
-            with an RP2040 on Windows, and is ignored on other operating systems
-        -f, --force
-            Force a device not in BOOTSEL mode but running compatible code to reset so the
-            command can be executed. After executing the command (unless the command itself is
-            a 'reboot') the device will be rebooted back to application mode
-        -F, --force-no-reboot
-            Force a device not in BOOTSEL mode but running compatible code to reset so the
-            command can be executed. After executing the command (unless the command itself is
-            a 'reboot') the device will be left connected and accessible to picotool, but
-            without the USB drive mounted
-        --bootsel-led <gpio>
-            Specify the GPIO for the BOOTSEL activity LED to flash (default none, ignored by
-            RP2350A-A2 in Arm mode) - only applicable if this command reboots the device to
-            BOOTSEL mode
-        --bootsel-led-active-low
-            The BOOTSEL activity LED is active low (ignored by RP2040 and RP2350-A4)
+        See "picotool help device-selection" for available options
     To target a file
         <filename>
             The file name
@@ -234,34 +211,7 @@ OPTIONS:
 
 TARGET SELECTION:
     To target one or more connected RP-series device(s) in BOOTSEL mode (the default)
-        --bus <bus>
-            Filter devices by USB bus number
-        --address <addr>
-            Filter devices by USB device address
-        --vid <vid>
-            Filter by vendor id
-        --pid <pid>
-            Filter by product id
-        --ser <ser>
-            Filter by serial number
-        --rp2040
-            Assume the device is an RP2040 - this is only required when using a custom vid/pid
-            with an RP2040 on Windows, and is ignored on other operating systems
-        -f, --force
-            Force a device not in BOOTSEL mode but running compatible code to reset so the
-            command can be executed. After executing the command (unless the command itself is
-            a 'reboot') the device will be rebooted back to application mode
-        -F, --force-no-reboot
-            Force a device not in BOOTSEL mode but running compatible code to reset so the
-            command can be executed. After executing the command (unless the command itself is
-            a 'reboot') the device will be left connected and accessible to picotool, but
-            without the USB drive mounted
-        --bootsel-led <gpio>
-            Specify the GPIO for the BOOTSEL activity LED to flash (default none, ignored by
-            RP2350A-A2 in Arm mode) - only applicable if this command reboots the device to
-            BOOTSEL mode
-        --bootsel-led-active-low
-            The BOOTSEL activity LED is active low (ignored by RP2040 and RP2350-A4)
+        See "picotool help device-selection" for available options
     To target a file
         <filename>
             The file name
@@ -349,34 +299,7 @@ OPTIONS:
         <offset>
             Load offset (memory address; default 0x10000000)
     Target device selection
-        --bus <bus>
-            Filter devices by USB bus number
-        --address <addr>
-            Filter devices by USB device address
-        --vid <vid>
-            Filter by vendor id
-        --pid <pid>
-            Filter by product id
-        --ser <ser>
-            Filter by serial number
-        --rp2040
-            Assume the device is an RP2040 - this is only required when using a custom vid/pid
-            with an RP2040 on Windows, and is ignored on other operating systems
-        -f, --force
-            Force a device not in BOOTSEL mode but running compatible code to reset so the
-            command can be executed. After executing the command (unless the command itself is
-            a 'reboot') the device will be rebooted back to application mode
-        -F, --force-no-reboot
-            Force a device not in BOOTSEL mode but running compatible code to reset so the
-            command can be executed. After executing the command (unless the command itself is
-            a 'reboot') the device will be left connected and accessible to picotool, but
-            without the USB drive mounted
-        --bootsel-led <gpio>
-            Specify the GPIO for the BOOTSEL activity LED to flash (default none, ignored by
-            RP2350A-A2 in Arm mode) - only applicable if this command reboots the device to
-            BOOTSEL mode
-        --bootsel-led-active-low
-            The BOOTSEL activity LED is active low (ignored by RP2040 and RP2350-A4)
+        See "picotool help device-selection" for available options
 ```
 
 e.g.
@@ -427,34 +350,7 @@ OPTIONS:
         -t <type>
             Specify file type (uf2 | elf | bin) explicitly, ignoring file extension
     Source device selection
-        --bus <bus>
-            Filter devices by USB bus number
-        --address <addr>
-            Filter devices by USB device address
-        --vid <vid>
-            Filter by vendor id
-        --pid <pid>
-            Filter by product id
-        --ser <ser>
-            Filter by serial number
-        --rp2040
-            Assume the device is an RP2040 - this is only required when using a custom vid/pid
-            with an RP2040 on Windows, and is ignored on other operating systems
-        -f, --force
-            Force a device not in BOOTSEL mode but running compatible code to reset so the
-            command can be executed. After executing the command (unless the command itself is
-            a 'reboot') the device will be rebooted back to application mode
-        -F, --force-no-reboot
-            Force a device not in BOOTSEL mode but running compatible code to reset so the
-            command can be executed. After executing the command (unless the command itself is
-            a 'reboot') the device will be left connected and accessible to picotool, but
-            without the USB drive mounted
-        --bootsel-led <gpio>
-            Specify the GPIO for the BOOTSEL activity LED to flash (default none, ignored by
-            RP2350A-A2 in Arm mode) - only applicable if this command reboots the device to
-            BOOTSEL mode
-        --bootsel-led-active-low
-            The BOOTSEL activity LED is active low (ignored by RP2040 and RP2350-A4)
+        See "picotool help device-selection" for available options
 ```
 
 e.g. first looking at what is on the device...
@@ -512,34 +408,7 @@ OPTIONS:
         <offset>
             Load offset (memory address; default 0x10000000)
     Target device selection
-        --bus <bus>
-            Filter devices by USB bus number
-        --address <addr>
-            Filter devices by USB device address
-        --vid <vid>
-            Filter by vendor id
-        --pid <pid>
-            Filter by product id
-        --ser <ser>
-            Filter by serial number
-        --rp2040
-            Assume the device is an RP2040 - this is only required when using a custom vid/pid
-            with an RP2040 on Windows, and is ignored on other operating systems
-        -f, --force
-            Force a device not in BOOTSEL mode but running compatible code to reset so the
-            command can be executed. After executing the command (unless the command itself is
-            a 'reboot') the device will be rebooted back to application mode
-        -F, --force-no-reboot
-            Force a device not in BOOTSEL mode but running compatible code to reset so the
-            command can be executed. After executing the command (unless the command itself is
-            a 'reboot') the device will be left connected and accessible to picotool, but
-            without the USB drive mounted
-        --bootsel-led <gpio>
-            Specify the GPIO for the BOOTSEL activity LED to flash (default none, ignored by
-            RP2350A-A2 in Arm mode) - only applicable if this command reboots the device to
-            BOOTSEL mode
-        --bootsel-led-active-low
-            The BOOTSEL activity LED is active low (ignored by RP2040 and RP2350-A4)
+        See "picotool help device-selection" for available options
 ```
 
 ## erase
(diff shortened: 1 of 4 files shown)


kilograham:
maybe worth renaming "device selection" at least the heading/docs (not the help option) to indicate that it has other stuff like reboot

main.cpp
```diff
@@ -723,7 +723,34 @@ auto device_selection =
...
+    ).min(0).doc_non_optional(true).collapse_synopsys("device-selection").help_topic("device-selection");
+
+struct help_topic {
+    string title;
+    string doc;
+    group options; // optional, leave blank for topics without options
+};
+
+string family_id_help_text() {
+    return string("A family ID identifies the target chip and/or image type of a UF2 file, and is "
```
lurch, line 735: Are family IDs specific to UF2 files, or are they also used as partition IDs?

main.cpp
```diff
@@ -1854,12 +1881,50 @@ int parse(const int argc, char **argv) {
...
+        for (const auto &major : options.contents.ordered_keys()) {
+            section_header(major.empty() ? "OPTIONS" : major);
+            bool first_opt = true;
+            for (const auto &minor : options.contents[major].ordered_keys()) {
+                fos.first_column(tab);
+                fos.hanging_indent(tab*2);
+                if (!minor.empty()) {
+                    fos << minor << "\n";
+                } else if (!first_opt) {
+                    fos << "Other\n";
```
lurch, line 1911: What does this "Other" indicate?

README.md
```diff
@@ -65,13 +65,17 @@ COMMANDS:
     link        Link multiple binaries into one block loop.
     bdev        Commands related to embedded block devices
 
+TOPICS:
+    device-selection   Options for Target Device Selection
+    family-ids         Valid Family IDs
+
 Use "picotool help <cmd>" for more info
```
lurch: Perhaps this should now be
```suggestion
Use "picotool help <cmd>" or "picotool help <topic>" for more info
```
?
(Or "picotool help <cmd|topic>" if you want to be more concise?)

main.cpp
will-v-pi: Note sure - this logic is copied from below, this change is just to put it into a lambda so it can be re-used in `print_help_topic`

README.md
```diff
@@ -1937,9 +1937,9 @@ purposes of binaries compiled using the [pico-sdk](https://github.com/raspberryp
 If your code has returned then rebooting with `-f/F` will not work - instead you can set the compile definition `PICO_ENTER_USB_BOOT_ON_EXIT`
 to reboot and be accessible to picotool once your code has finished execution, for example with
 `target_compile_definitions(<yourTargetName> PRIVATE PICO_ENTER_USB_BOOT_ON_EXIT=1)`
-- Uses stdio_usb -
-If your binary calls `stdio_init_all()` and you have `pico_enable_stdio_usb(<yourTargetName> 1)` in your CMakeLists.txt file then you meet
-this requirement (see the [hello_usb](https://github.com/raspberrypi/pico-examples/tree/master/hello_world/usb) example)
+- Uses stdio_usb or the pico_usb_reset library -
```
lurch, line 1949: Maybe it's worth splitting these into two separate bullet-points?

main.cpp
```diff
@@ -741,8 +741,11 @@ string family_id_help_text() {
 std::map<string, help_topic> help_topics {
 #if HAS_LIBUSB
     { "device-selection", {
-        "Options for Target Device Selection",
-        "Options for selecting or filtering the target RP-series device(s). These options are accepted by most commands which target a device.",
+        "Options for Target Device Selection and Rebooting",
+        "Options for selecting or filtering the target RP-series device(s), including rebooting devices to BOOTSEL mode first. "
+        "These options are accepted by most commands which target a device.\n\n"
+        "For the `-f/F` options, compatible code is defined as code which uses stdio_usb or the pico_usb_reset library, and which is still running. "
```
lurch, line 747: I'm not sure if the "compatible code is defined as" is useful in this help-text? Perhaps it could be shortened to
"For the `-f/F` options, the program must be using the stdio_usb or pico_usb_reset libraries, and still be running."

EDIT: Also, I dunno if "the `-f/F` options" would be better as "the `-f/-F` options" ? :thinking:

main.cpp
will-v-pi: They were specific to UF2 files, but partitions can also be tagged with the UF2 family IDs which they accept.

`--family` is currently used by:

- `save`, `partition create`, `uf2 convert`, `uf2 combine` (to set the ID of the output UF2 file)
- `load` (to load it where a UF2 file with that ID would go)
- `partition info` (to see where a UF2 file with that ID would go).

Partition IDs are separate 64-bit IDs which partitions can optionally be assigned, e.g. to mark a partition as a block device (`BLOCK_DEVICE_DEFAULT_PARTITION_ID`) or mark it as containing Wi-Fi firmware (`CYW43_FIRMWARE_PARTITION_ID`)

main.cpp
will-v-pi: The `-f/F` options below refer to "compatible code", hence including "compatible code is defined as" here

README.md
will-v-pi: `stdio_usb` uses the `pico_usb_reset` library behind the scenes, so I think keeping them in a single bullet point works better? Note that the `dev_multi_cdc` example doesn't link `pico_usb_reset` directly, it uses `stdio_usb` and links `tinyusb_device`

README.md
will-v-pi: I think the more verbose option is cleaner - will add that

will-v-pi:
> maybe worth renaming "device selection" at least the heading/docs (not the help option) to indicate that it has other stuff like reboot

I have updated it to "Options for Target Device Selection and Rebooting"

main.cpp
lurch: Ahhh, in that case I was probably just confusing (mis-remembering) the `cyw43-firmware` family ID and the `CYW43_FIRMWARE_PARTITION_ID` partition ID :upside_down_face:

README.md
```diff
@@ -1933,6 +1919,7 @@ enum {
     BINARY_INFO_BLOCK_DEV_FLAG_PT_NONE = 3 << 4, // no partition table
 };
 ```
+For more information see the Pico SDK, where this macro and flags are defined.
```
lurch, line 1939: Is this `bi_block_device` macro ideally something that ought to appear at e.g. https://www.raspberrypi.com/documentation/pico-sdk/runtime.html#macros-2 too?

will-v-pi: That would probably be good, along with all the other macros from `pico/binary_info/code.h` which don't begin with `__`. Currently it looks like all the structs are documented there instead of the macros, which is incorrect as users should use the macros not the structs

lurch: Cool, I just made an issue for this: https://github.com/raspberrypi/picotool/issues/345

gen_help_txt.sh
```diff
@@ -97,17 +97,57 @@ if [ ${#missing_commands[@]} -ne 0 ]; then
...
+    if [ ! -z "$topic" ]; then
+        if ! grep -q "\$ picotool help $topic" tmp/README.md; then
+            missing_topics+=("$topic")
+            echo "Missing help text section for topic: $topic"
+        fi
+    fi
+done <<< "$topics"
+
+# Same as for missing commands above - still fails the job later, this is just to
+# provide the output to copy into the right place
```
lurch, line 114: Does the earlier "Adding missing help text sections to end of file..." section need to be modified so that any missing command-help sections get added _before_ the topic-help sections? :thinking: 

EDIT: Also, should there be a `check_topics_order()`, analogous to `check_cmd_order()`?

EDIT2: And is there a reason why this script does both `perl -i -pe ...` and `perl -i -0777 -pe ...`?

README.md
```diff
@@ -2401,27 +1900,13 @@ quotes, newlines etc you may have better luck defining via bi_decl in the code.
 ### Block devices
 
 MicroPython and CircuitPython, eventually the SDK and others may support one or more storage devices in flash. We already
-have macros to define these although picotool doesn't do anything with them yet... but backup/restore/file copy and even fuse mount
-in the future might be interesting.
+have macros to define these, and they can be accessed with the [`bdev`](#bdev) commands. In the future backup/restore and even fuse mount might be interesting additions.
```
```suggestion
have macros to define these, and they can be accessed with the [`bdev`](#bdev) commands. In the future, backup/restore and even fuse mount might be interesting additions.
```

lurch:
Perhaps the "Show general help or help for a specific command" now ought to say "Show general help, or help for a specific command or topic".

README.md
will-v-pi: Thanks - I've moved the issue over to pico-sdk

gen_help_txt.sh
will-v-pi: > Does the earlier "Adding missing help text sections to end of file..." section need to be modified so that any missing command-help sections get added _before_ the topic-help sections? 🤔

No, because the missing sections are only added to make copying them into place easier - they should not be left where they are, they're just placed at the end for simplicity

> EDIT: Also, should there be a `check_topics_order()`, analogous to `check_cmd_order()`?

Yes, I will add that

> EDIT2: And is there a reason why this script does both `perl -i -pe ...` and `perl -i -0777 -pe ...`?

Not sure - I will standardise them both to the `-0777` variant

README.md
lurch: Thanks :slightly_smiling_face:

main.cpp
lurch: You can see "Other" being used at https://github.com/raspberrypi/picotool#save - which curiously is the only place it appears!

main.cpp
```diff
@@ -2090,6 +2090,12 @@ int parse(const int argc, char **argv) {
 
     auto args = cli::make_args(argc, argv);
 
+    // Check if the only argument is --version, if so replace it with version
+    if (args[0] == "--version" && args.size() == 1) {
+        printf("WARNING: Use \"picotool version\" instead of \"picotool --version\"\n");
```
lurch, line 2095: Should ERROR: and WARNING: strings go to stderr instead of stdout?

will-v-pi: No, everything goes to stdout - I don't want to change that as I have scripts that rely on that behaviour, and others probably do too, so that would be quite a breaking change.

lurch: Okay :+1:

README.md
```diff
@@ -44,41 +44,98 @@ SYNOPSIS:
...
-    reboot      Reboot the device
+    reboot      Reboot the device.
     seal        Add final metadata to a binary, optionally including a hash and/or signature.
     encrypt     Encrypt the program.
-    partition   Commands related to RP2350 Partition Tables
-    uf2         Commands related to UF2 creation and status
-    otp         Commands related to the RP2350 OTP (One-Time-Programmable) Memory
+    partition   Commands related to RP2350 Partition Tables.
+    uf2         Commands related to UF2 creation and status.
+    otp         Commands related to the RP2350 OTP (One-Time-Programmable) Memory.
```
```suggestion
    otp         Commands related to the RP2350 OTP (One-Time-Programmable) memory.
```

README.md
```diff
@@ -44,41 +44,98 @@ SYNOPSIS:
...
     coprodis    Post-process coprocessor instructions in disassembly files.
     link        Link multiple binaries into one block loop.
-    bdev        Commands related to embedded block devices
+    bdev        Commands related to embedded block devices.
 
-Use "picotool help <cmd>" for more info
+TOPICS:
+    device-selection   Options for Target Device Selection and Rebooting
```
lurch, line 69: The selection of Capital letters here feels a Bit Random? :thinking:

README.md
```diff
@@ -44,41 +44,98 @@ SYNOPSIS:
...
+        -s, --semantic
+            Output semantic version number only
+        <version>
+            Check compatibility with version
+```
+
+The normal output is this format:
+```text
+$ picotool version
+picotool v2.3.0 (Linux, GNU-12.2.0, Release)
```
lurch: _Probably_ not worth it, but I guess all these v2.3.0 versions could be updated to v2.3.1 ?

README.md
```diff
@@ -44,41 +44,98 @@ SYNOPSIS:
...
+2.3.0
+```
+
+When checking compatibility with a compatible version (e.g. 2.2.0 with version 2.3.0):
+```text
+$ picotool version -s 2.2.0
+2.3.0
+```
+
+Or with an incompatible version (e.g. 2.4.0 with version 2.3.0), it will exit with error code `ERROR_INCOMPATIBLE` (-3):
```
lurch: It earlier says "to check a newer `picotool` is backwards compatible with an older version", so perhaps it doesn't make sense to try checking against a not-yet-released picotool version, and would be better to check against an _older_ version that picotool 2.3.0 **isn't** backwards-compatible with?

README.md
```diff
@@ -225,47 +255,18 @@ SYNOPSIS:
     picotool config [-s <key> <value>] [-g <group>] <filename> [-t <type>]
 
 OPTIONS:
-        <key>
-            Variable name
-        <value>
-            New value
-        -g <group>
+        -s, --set <key> <value>
+            Set config variable name to new value
```
lurch, line 259: Outside the scope of _this_ PR, but if `picotool config` allows you to **set** the value of a particular config key, it'd probably also be useful if it allowed you to **get** the value of a single config key?

README.md
will-v-pi: It’s just title case

README.md
will-v-pi: I’ll merge in `develop` so they get updated, otherwise the CI readme check will fail

README.md
```diff
@@ -498,48 +437,16 @@ OPTIONS:
...
-            Compare a sub range of memory only
-        <from>
-            The lower address bound in hex
-        <to>
-            The upper address bound in hex
-        -o, --offset
-            Specify the load address when comparing with a BIN file
-        <offset>
-            Load offset (memory address; default 0x10000000)
+        -r, --range <from> <to>
```
lurch, line 440: Why does the help-text for the `erase` command below not also use this new single-line `-r, --range <from> <to>` format?

README.md
```diff
@@ -805,10 +656,9 @@ OPTIONS:
             Use ~180MHz ROSC configuration for embedded bootloader
         --use-mbedtls
             Use MbedTLS implementation of embedded bootloader (faster but less secure)
-        --otp-key-page
-            Specify the OTP page storing the AES key (IV salt is stored on the next page)
-        <page>
-            OTP page (default 29)
+        --otp-key-page <page>
```
lurch, line 650: Does this command fail if given an `otp-key-page` but not an `otp` JSON filename? (And similarly does it fail if given an `otp` JSON filename but not an `otp-key-page`?)

And presumably the `--sign` command fails if not given a `<signing_key>`?

Hmm, is the `<signing_key>` option of the `encrypt` command the same thing as the `<key>` option of the `seal` command? (If they are, _perhaps_ it makes sense to rename `seal`'s option for consistency?)

README.md
```diff
@@ -935,26 +775,22 @@ OPTIONS:
...
+            Specify the family id for UF2 file output (default absolute)
     embed partition table into bootloader ELF
         <bootloader>
             The file name
-        -t <type>
+        -t, --type <type>
             Specify file type (elf) explicitly, ignoring file extension
     Partition Table Options
         --sign <keyfile>
             The file name
```
lurch: Should "The file name" here be "Signing Key file (.pem)" ?

Hmm, should the `<bootloader>` option's "The file name" text be clarified too?

README.md
will-v-pi: Oh actually, this is manually generated - I'll update them while this PR isn't merged, but won't update them going forwards

README.md
will-v-pi: There isn't really a good example for that - it could check with a pre-2.0.0 picotool (e.g. 1.1.2), but this version check is primarily intended for the SDK to check for a compatible `picotool`, which only applies since SDK 2.0.0.

The better demo here would be asking version 2.2.0 if it's compatible with version 2.3.0, so I'll switch it to that

README.md
will-v-pi: It hasn't been changed in `main.cpp` - nor has `save`

will-v-pi: Looks like there are still some issues doing that here, as these are `|` together, not `+` - I will see if I can get it working

README.md
will-v-pi: > Does this command fail if given an `otp-key-page` but not an `otp` JSON filename? (And similarly does it fail if given an `otp` JSON filename but not an `otp-key-page`?)

No - both of those are entirely valid use-cases (the key page is in the binary, not just the JSON file, and has a default value when not provided)

> And presumably the `--sign` command fails if not given a `<signing_key>`?

Yes, that does fail with `missing key file for signing after encryption`

> Hmm, is the `<signing_key>` option of the `encrypt` command the same thing as the `<key>` option of the `seal` command? (If they are, _perhaps_ it makes sense to rename `seal`'s option for consistency?)

They are the same, but `seal` only takes 1 key, whereas `encrypt` takes a signing key and an encryption key, hence having different names

README.md
will-v-pi: Have fixed up the keyfile, the bootloader file is documented above as "Embed partition table into bootloader ELF"

PR #346, opened 2026-07-16
Get model from family, if '--platform' is missing

### LINKED ISSUE #339: picotool 2.3.0 fails in uf2 convert for images that do not start at FLASH_START
Converting a valid RP2350 flash ELF whose lowest segment doesn't start at the absolute FLASH_START (app built to run from a
bootloader),  causes picotool to silently assume the chip is RP2040 instead of RP2350. This causes a valid RP2350 addresses to be rejected with:
```
    ERROR: Memory segment 20081000->20081800 is outside of valid address range for device
```
The image itself is correct. This only happens when picotool has to guess the chip, no --platform was given and no IMAGE_DEF block was found.

tzstoyanov:
If the user does not use the `--platform` parameter to specify the target platform for uf2 convert command, get the model from the target family. The current logic tries to guess the platform by analyzing the target binary file. It works OK in most regular cases, but fails in some weird cases where the image is not located at absolute FLASH_START offset.

fixes #339

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -4558,10 +4558,17 @@ uint32_t get_family_id(uint8_t file_idx) {
 }
 
 model_t get_model(uint8_t file_idx) {
-    model_t model;
+    model_t model = nullptr;
     if (settings.model) {
         model = settings.model;
-    } else {
+    } else if (settings.family_id) {
+        auto model_f = model_from_family(settings.family_id);
+        if (model_f->chip() != unknown) {
+            settings.model = model_f;
+            model = settings.model;
+        }
+    }
+    if (model == nullptr) {
         auto file_access = get_file_memory_access(file_idx);
         model = get_access_model(file_access);
     }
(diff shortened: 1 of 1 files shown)


main.cpp
```diff
@@ -4558,10 +4558,17 @@ uint32_t get_family_id(uint8_t file_idx) {
...
-    model_t model;
+    model_t model = nullptr;
     if (settings.model) {
         model = settings.model;
-    } else {
+    } else if (settings.family_id) {
+        auto model_f = model_from_family(settings.family_id);
+        if (model_f->chip() != unknown) {
+            settings.model = model_f;
+            model = settings.model;
```
will-v-pi: I think I prefer `model_guess` to `model_f` as the variable name - also I don't think this should set `settings.model`

```suggestion
        auto model_guess = model_from_family(settings.family_id);
        if (model_guess->chip() != unknown) {
            model = model_guess;
```

will-v-pi (approved):

PR #349, opened 2026-07-17
Add load support for universal binaries

### LINKED ISSUE #344: Picotool load combined uf2 ignoring family ID
According to https://github.com/microsoft/uf2#family-id , the family ID is there to allow multiple uf2 files combined into one for different targets.
(and also some comments in https://github.com/raspberrypi/picotool/pull/313 )
This works well with the RP2040 bootloader, when copying the uf2 ( `cp test2.uf2 /media/arendjan/RPI-RP2/` ) then my board runs the correct firmware, either first or second part of the combined uf2.

When using `picotool load -f combined.uf2`, I get:
```
Tracking device serial number E6625C48934A7E32 for reboot
The device was asked to reboot into BOOTSEL mode so the command can be executed.

WARNING: Multiple family IDs in a single UF2 file - only using first one
WARNING: Multiple family IDs in a single UF2 file - only using first one
Loading into Flash:   [==============================]  100%

The device was asked to reboot back into application mode.
```

and the board either boots the code (when first part of uf2 is for it) or reconnects in bootloader mode (when correct blocks are in second part).

Using family flags ( `--family rp2040` ) doesn't help.

I'd assume picotool would only upload the firmware for the connected device (or the bootloader would filter it).

(picotool build from current main)

# RP2350 also not working:
UF2 correct part is second one:
```
Tracking device serial number DA72B1529E0666F5 for reboot
The device was asked to reboot into BOOTSEL mode so the command can be executed.

WARNING: Multiple family IDs in a single UF2 file - only using first one
Family ID 'rp2040' cannot be downloaded anywhere
ERROR: This file cannot be loaded onto a device with no partition table
```

UF2 correct part is first one:
```
WARNING: Multiple family IDs in a single UF2 file - only using first one
Family ID 'rp2350-arm-s' can be downloaded in absolute space:
  00000000->02000000
WARNING: Multiple family IDs in a single UF2 file - only using first one
Loading into Flash:   [==============================]  100%
```

will-v-pi:
This adds support to `picotool load` for loading universal binaries, and other multi-family UF2s

The existing `--family` flag can be used to select which family to load from the UF2 - otherwise it picks using the partition table, or based on the supported families for the device for RP2040 or SRAM binaries (no partition table)

Also fixes `picotool info` and other command support for 2.3.0 universal binaries, which have rearranged blocks that picotool was handling incorrectly

Fixes #344

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,7 +314,7 @@ SYNOPSIS:
 OPTIONS:
     Post load actions
         --ignore-partitions
-            When writing flash data, ignore the partition table and write to absolute space
+            Ignore the partition table, or whether the family ID is compatible with the device
         --family
             Specify the family ID of the file to load
         <family_id>
--- a/README.md
+++ b/README.md
@@ -312,9 +312,9 @@ SYNOPSIS:
                 [-v] [-x] <filename> [-t <type>] [-o <offset>] [device-selection]
 
 OPTIONS:
-    Post load actions
+    Load options
         --ignore-partitions
-            Ignore the partition table, or whether the family ID is compatible with the device
+            Ignore the partition table, and ignore family ID compatibility
         --family
             Specify the family ID of the file to load
         <family_id>
--- a/main.cpp
+++ b/main.cpp
@@ -1094,7 +1094,7 @@ struct load_command : public cmd {
     group get_cli() override {
         return (
             (
-                option("--ignore-partitions").set(settings.load.ignore_pt) % "When writing flash data, ignore the partition table and write to absolute space" +
+                option("--ignore-partitions").set(settings.load.ignore_pt) % "Ignore the partition table, or whether the family ID is compatible with the device" +
                 (option("--family") % "Specify the family ID of the file to load" &
                         family_id("family_id").set(settings.family_id) % "family ID to use for load").force_expand_help(true) +
                 (option('p', "--partition") % "Specify the partition to load into" &
--- a/main.cpp
+++ b/main.cpp
@@ -5413,7 +5413,31 @@ bool load_guts(picoboot::connection con, iostream_memory_access &file_access) {
 bool load_command::execute(device_map &devices) {
     auto con = get_single_bootsel_device_connection(devices);
     picoboot_memory_access raw_access(con);
-    auto tmp_file_access = get_file_memory_access(0);
+    uint32_t next_id = 0;
+    vector<uint32_t> available_family_ids;
+    uint32_t file_family_id = 0;
+    uint32_t override_family_id = settings.family_id;
+    auto tmp_file_access = get_file_memory_access(0, false, &next_id);
+    if (next_id) {
+        // UF2 file with multiple family IDs
+        settings.family_id = 0;
+        next_id = get_family_id(0);
+        while (next_id) {
+            available_family_ids.push_back(next_id);
+            auto tmp_access = get_file_memory_access(0, false, &next_id);
+        }
+        if (override_family_id) {
+            if (std::find(available_family_ids.begin(), available_family_ids.end(), override_family_id) == available_family_ids.end()) {
+                fos << "WARNING: Requested family ID " << family_name(override_family_id) << " not found in UF2 file, ";
+                fos << "so treating first family ID found in the UF2 file (" << family_name(available_family_ids[0]) << ") as the requested one\n";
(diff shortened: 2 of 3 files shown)


main.cpp
```diff
@@ -3306,12 +3306,32 @@ uint32_t build_rmap_uf2(std::shared_ptr<std::iostream>file, range_map<size_t>& r
...
+    #if SUPPORT_RP2350_A2
+            && !check_abs_block(block)
+    #endif
+        ) {
+            // new family ID found
+            family_ids.push_back(block.file_size);
+            DEBUG_LOG("New family ID %s\n", family_name(block.file_size).c_str());
+            if (
+                std::find(family_ids.begin(), family_ids.end(), family_id) != family_ids.end()
+                && std::find(family_ids.begin(), family_ids.end(), family_id) + 1 == std::find(family_ids.begin(), family_ids.end(), block.file_size)
```
lurch: Maybe it makes sense to cache the values of `std::find(family_ids.begin(), family_ids.end(), block.file_size)` and `std::find(family_ids.begin(), family_ids.end(), family_id)`, rather than computing them multiple times?
The runtime reduction would probably be fairly minimal, but it might make the code a bit more readable?

main.cpp
```diff
@@ -5413,7 +5424,49 @@ bool load_guts(picoboot::connection con, iostream_memory_access &file_access) {
...
+        next_id = get_family_id(0);
+        while (next_id) {
+            available_family_ids.push_back(next_id);
+            auto tmp_access = get_file_memory_access(0, false, &next_id);
+        }
+        if (override_family_id) {
+            if (std::find(available_family_ids.begin(), available_family_ids.end(), override_family_id) == available_family_ids.end()) {
+                fos << "WARNING: Requested family ID " << family_name(override_family_id) << " not found in UF2 file, ";
+                fos << "so loading the first family ID found in the UF2 file (" << family_name(available_family_ids[0]) << "), ";
+                fos << "into where family ID " << family_name(override_family_id) << " would go\n";
```
lurch, line 5446: Should the word "partition" appear between "into" and "where"?

will-v-pi: This could also apply on RP2040, or to SRAM binaries - although in those cases it’ll just error if `--family` isn’t compatible with the device, rather than placing it anywhere, so there might be a better way to word this? Basically it loads it as if it had the family specified by `--family`

lurch: Perhaps "into the location where family ID XXXX would normally go." ?

main.cpp
will-v-pi: The `block.file_size` one can't be cached as it changes between the two checks (`block.file_size` is added to the vector), and the `family_id` one is only performed twice per family ID in a file so probably not worth it

main.cpp
```suggestion
                fos << "into the location where family ID " << family_name(override_family_id) << " would normally go\n";
```

README.md
```diff
@@ -314,7 +314,7 @@ SYNOPSIS:
 OPTIONS:
     Post load actions
```
lurch: I guess some of these aren't post-load options :wink:

README.md
```diff
@@ -314,7 +314,7 @@ SYNOPSIS:
...
         --ignore-partitions
-            When writing flash data, ignore the partition table and write to absolute space
+            Ignore the partition table, or whether the family ID is compatible with the device
```
lurch: Would there ever be a use-case for this being a separate `--ignore-family-ids` option rather than shoe-horning this functionality into `--ignore-partitions`?

If not, perhaps this wording is slightly clearer?
```suggestion
            Ignore the partition table, and ignore family ID compatibility
```

will-v-pi: I don't think it's worth adding a separate argument, as they'd be mutually exclusive - the family ID ignoring is for binaries targetting SRAM only (or all binaries on RP2040), whereas the partition ignoring is for binaries targeting Flash

will-v-pi: ~~I think I'll go for this, as it's always family ID compatibility, just whether it's with the partition table or the SRAM accepted families~~
```
Ignore family ID compatibility with the partition table or the device
```

EDIT: actually, will stick with your suggestion, as it is also ignoring the partition table when placing the binary, not just when deciding compatibility

lurch:
I've not tested the code, but thanks for making the wording-tweaks I suggested :+1:

PR #350, opened 2026-07-22
Add support for rebooting debugprobes, and list command

### LINKED ISSUE #26: Support "list" command
I wonder if a `pictool list` command might be useful? Consider the following scenario:
```
% lsusb | grep 2e8a                                       
Bus 020 Device 012: ID 2e8a:000a 2e8a Pico  Serial: 000000000000
Bus 020 Device 009: ID 2e8a:0005 2e8a Board in FS mode  Serial: 000000000000
% ./picotool info

but:

Device at bus 20, address 9 appears to be a RP2040 MicroPython device not in BOOTSEL mode.
Device at bus 20, address 12 appears to be a RP2040 device with a USB serial connection, so consider -f or -F.
```
if I then plug in a Pico with a blank Flash chip (wiped using `flash_nuke.uf2`) I get:
```
% lsusb | grep 2e8a
Bus 020 Device 012: ID 2e8a:000a 2e8a Pico  Serial: 000000000000
Bus 020 Device 009: ID 2e8a:0005 2e8a Board in FS mode  Serial: 000000000000
Bus 020 Device 013: ID 2e8a:0003 2e8a RP2 Boot  Serial: E0C912D24340
% ./picotool info
Program Information
 none
```
In this latter case it would be nice if e.g. a `./picotool list` could tell me:
```
Device at bus 20, address 9 appears to be a RP2040 MicroPython device not in BOOTSEL mode.
Device at bus 20, address 12 appears to be a RP2040 device with a USB serial connection, so consider -f or -F.
Device at bus 20, address 13 appears to be a RP2040 device in BOOTSEL mode.
```
For bonus points, it could even include the other VID/PID combos from https://github.com/raspberrypi/usb-pid ? :wink:

will-v-pi:
Uses the `--debugprobe` argument, to explicitly target only running debugprobes, and ignore any BOOTSEL devices

Will work with the current master branch, and the next firmware release

Also refactors the usb reset support to be clearer what is going on, and add `--ser` filtering for custom vid/pid which was missing

Also adds `--only-force` argument for the same behaviour, in case that functionality is useful

Also adds `picotool list` command, which just prints the no device found  help text, including BOOTSEL devices

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ SYNOPSIS:
     picotool erase -p <partition> [device-selection]
     picotool erase -r <from> <to> [device-selection]
     picotool reboot [-a] [-u] [-g <partition>] [-c <cpu>] [device-selection]
+    picotool list 
     picotool seal [--quiet] [--verbose] [--hash] [--sign] [--clear] [--pin-xip-sram]
                 [--no-squash] <infile> [-t <type>] [-o <offset>] <outfile> [-t <type>] [<key>]
                 [<otp>] [--major <major>] [--minor <minor>] [--rollback <rollback> [<rows>..]]
@@ -56,6 +57,7 @@ COMMANDS:
     verify      Check that the device contents match those in the file.
     erase       Erase the program / memory stored in flash on the device.
     reboot      Reboot the device
+    list        List all connected RP-series devices
     seal        Add final metadata to a binary, optionally including a hash and/or signature.
     encrypt     Encrypt the program.
     partition   Commands related to RP2350 Partition Tables
@@ -71,7 +73,7 @@ Use "picotool help <cmd>" for more info
 Note commands that aren't acting on files require a device in BOOTSEL mode to be connected.
 
 ## Links to documentation for `picotool` commands
-[`info`](#info) [`config`](#config) [`load`](#load) [`save`](#save) [`verify`](#verify) [`erase`](#erase) [`reboot`](#reboot) [`seal`](#seal) [`encrypt`](#encrypt) [`partition`](#partition) [`uf2`](#uf2) [`otp`](#otp) [`coprodis`](#coprodis) [`link`](#link) [`bdev`](#bdev)
+[`info`](#info) [`config`](#config) [`load`](#load) [`save`](#save) [`verify`](#verify) [`erase`](#erase) [`reboot`](#reboot) [`list`](#list) [`seal`](#seal) [`encrypt`](#encrypt) [`partition`](#partition) [`uf2`](#uf2) [`otp`](#otp) [`coprodis`](#coprodis) [`link`](#link) [`bdev`](#bdev)
 
 ## Building & Installing
 
@@ -736,6 +738,60 @@ OPTIONS:
             BOOTSEL mode
         --bootsel-led-active-low
             The BOOTSEL activity LED is active low (ignored by RP2040 and RP2350-A4)
+```
+
+## list
+
+`list` allows you to list all connected RP-series devices
+
+```text
+$ picotool help list
+LIST:
+    List all connected RP-series devices
+
+SYNOPSIS:
+    picotool list 
+```
+
+For example, with a DebugProbe and an RP2350 connected:
+
+```text
+$ picotool list
+Detected 2 RP-series devices:
+
+  RP2350 device at bus 1, address 25 appears to have a USB reset interface, so consider -f (or
+      -F) to force reboot in order to run the command.
+
+  RP2040 device at bus 1, address 27 appears to be an RP-series DebugProbe device not in
+      BOOTSEL mode, but with a USB reset interface, so consider --debugprobe to force reboot it
+      in order to run the command.
+
+```
+
+Then after running `picotool reboot -f -u`:
+
+```text
+$ picotool list
+Detected 2 RP-series devices:
+
+  RP2350 device at bus 1, address 29 appears to be an RP-series device in BOOTSEL mode.
+
+  RP2040 device at bus 1, address 27 appears to be an RP-series DebugProbe device not in
+      BOOTSEL mode, but with a USB reset interface, so consider --debugprobe to force reboot it
+      in order to run the command.
+
+```
+
+And also running `picotool reboot --debugprobe -u`:
+
+```text
+$ picotool list
+Detected 2 RP-series devices:
+
+  RP2040 device at bus 1, address 30 appears to be an RP-series device in BOOTSEL mode.
+
+  RP2350 device at bus 1, address 29 appears to be an RP-series device in BOOTSEL mode.
+
 ```
 
 ## seal
--- a/README.md
+++ b/README.md
@@ -759,12 +759,11 @@ For example, with a DebugProbe and an RP2350 connected:
 $ picotool list
 Detected 2 RP-series devices:
 
-  RP2350 device at bus 1, address 25 appears to have a USB reset interface, so consider -f (or
+  RP2350 device at bus 1, address 31 appears to have a USB reset interface, so consider -f (or
       -F) to force reboot in order to run the command.
 
-  RP2040 device at bus 1, address 27 appears to be an RP-series DebugProbe device not in
-      BOOTSEL mode, but with a USB reset interface, so consider --debugprobe to force reboot it
-      in order to run the command.
+  RP2040 device at bus 1, address 32 appears to be a DebugProbe with a USB reset interface, so
+      consider --debugprobe to force reboot it in order to run the command.
 
 ```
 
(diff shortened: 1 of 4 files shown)


main.cpp
```diff
@@ -708,12 +710,14 @@ auto device_selection =
...
         (option("--address") & integer("addr").min_value(1).max_value(127).set(settings.address)
             .if_missing([] { return "missing address"; })) % "Filter devices by USB device address" +
-        (option("--vid") & integer("vid").set(settings.vid).if_missing([] { return "missing vid"; })) % "Filter by vendor id" +
+        (option("--vid") & integer("vid").set(settings.vid).min(1).if_missing([] { return "missing vid"; })) % "Filter by vendor id" +
         (option("--pid") & integer("pid").set(settings.pid)) % "Filter by product id" +
         (option("--ser") & value("ser").set(settings.ser)) % "Filter by serial number"
         + option("--rp2040").set(settings.force_rp2040) % "Assume the device is an RP2040 - this is only required when using a custom vid/pid with an RP2040 on Windows, and is ignored on other operating systems"
+        + option("--debugprobe").set(settings.target_debugprobe) % "Force a DebugProbe device version >2.3.1 to reset so the command can be executed. After executing the command (unless the command itself is a 'reboot') the device will be rebooted back to application mode"
```
lurch: It's the firmware version that's important; IMHO the existing wording made it sound like it was the hardware version that mattered.
```suggestion
        + option("--debugprobe").set(settings.target_debugprobe) % "Force a DebugProbe device using firmware version >2.3.1 to reset so the command can be executed. After executing the command (unless the command itself is a 'reboot') the device will be rebooted back to application mode"
```

main.cpp
```diff
@@ -708,12 +710,14 @@ auto device_selection =
...
         + option('f', "--force").set(settings.force) % "Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the command (unless the command itself is a 'reboot') the device will be rebooted back to application mode" +
                 option('F', "--force-no-reboot").set(settings.force_no_reboot) % "Force a device not in BOOTSEL mode but running compatible code to reset so the command can be executed. After executing the command (unless the command itself is a 'reboot') the device will be left connected and accessible to picotool, but without the USB drive mounted"
+        + option("--only-force").set(settings.only_force) % "Same as --force, but will only target devices not currently in BOOTSEL mode"
```
lurch: Hmmm? The help-text for `--force` says "Force a device not in BOOTSEL mode..."

picoboot_connection/picoboot_connection.c
```diff
@@ -66,6 +66,20 @@ enum picoboot_device_result picoboot_open_device(libusb_device *device, libusb_d
     definitely_exclusive = false;
     *dev_handle = NULL;
     *chip = unknown;
+    bool custom_vid_pid = !(vid < 0 && pid < 0);
+    bool checking_for_debugprobe = false;
+    if (vid <= 0 || vid == VENDOR_ID_RASPBERRY_PI) { // no filtering is not custom_vid_pid
```
lurch:
```suggestion
    if (vid <= 0 || vid == VENDOR_ID_RASPBERRY_PI) { // no filtering if not custom_vid_pid
```
?

picoboot_connection/picoboot_connection.h
```diff
@@ -33,11 +35,16 @@ enum picoboot_device_result {
...
+    dr_vidpid_circuitpython,
     dr_vidpid_picoprobe,
+    dr_vidpid_debugprobe,
+    dr_vidpid_debugprobe_cant_connect,
+    dr_vidpid_debugprobe_no_reset,
     dr_vidpid_unknown,
     dr_error,
-    dr_vidpid_stdio_usb,
-    dr_vidpid_stdio_usb_cant_connect,
+    dr_vidpid_usb_reset,
```
lurch: If this is being renamed from `dr_vidpid_stdio_usb` to `dr_vidpid_usb_reset`, then should e.g. `PRODUCT_ID_STDIO_USB` also be renamed to `PRODUCT_ID_USB_RESET`? :thinking: 
Or is there still a distinction between "stdio usb with no reset interface" and "stdio usb with a reset interface"?

main.cpp
```diff
@@ -10292,25 +10320,53 @@ int main(int argc, char **argv) {
...
                                     " appears to be an RP-series PicoProbe device not in BOOTSEL mode.");
                             printer(dr_vidpid_micropython,
                                     " appears to be an RP-series MicroPython device not in BOOTSEL mode.");
+                            printer(dr_vidpid_circuitpython,
+                                    " appears to be an RP-series CircuitPython device not in BOOTSEL mode.");
+                            printer(dr_vidpid_debugprobe,
+                                    " appears to be an RP-series DebugProbe device not in BOOTSEL mode, but with a USB reset interface, so consider --debugprobe to force reboot it in order to run the command.");
+                            if (settings.target_debugprobe) {
+                                printer(dr_vidpid_debugprobe_no_reset,
+                                        " appears to be an RP-series DebugProbe device not in BOOTSEL mode, but it does not have a USB reset interface so must be manually entered into BOOTSEL mode.");
```
lurch: I'm not sure if the "must be manually entered into BOOTSEL mode" phrasing here sounds correct :confused: But the best alternative I could come up with is "must be manually put into BOOTSEL mode".

README.md
```diff
@@ -130,6 +130,10 @@ TARGET SELECTION:
         --rp2040
             Assume the device is an RP2040 - this is only required when using a custom vid/pid
             with an RP2040 on Windows, and is ignored on other operating systems
+        --debugprobe
```
lurch, line 135: I guess that once #342 has been merged, these `--debugprobe` and `--only-force` help sections will become part of the device-selection help topic?

main.cpp
lurch: And of course this raises the question: how does `picotool` treat DebugProbe devices running firmwares <2.3.1 ?

main.cpp
will-v-pi: Yes, but `-f` will disengage if it detects a device in BOOTSEL mode and use that instead - this allows always passing `-f` to commands, and they will work if the device is running or in BOOTSEL mode

This new option ignores devices already in BOOTSEL mode

picoboot_connection/picoboot_connection.h
will-v-pi: The product ID comes from stdio_usb - the reset interface can be valid for any product ID, hence separating it out

You can also disable the reset interface when using stdio_usb, which is a case the previous code didn’t handle - I need to check this code handles it clearly

README.md
will-v-pi: Yep, that was part of my reason for doing the topics, because I knew I wanted to add the debugprobe argument

picoboot_connection/picoboot_connection.h
lurch: > The product ID comes from stdio_usb - the reset interface can be valid for any product ID, hence separating it out

Okay, so in that case does `dr_vidpid_usb_reset` mean "This is a device with the `PRODUCT_ID_STDIO_USB` pid **and** it has the reset interface enabled"? And does `dr_vidpid_usb_reset_cant_connect` similarly mean "This is a device with the `PRODUCT_ID_STDIO_USB` pid _without_ the reset interface enabled"? Or am I still misunderstanding things?

main.cpp
lurch: Hmmm, so I guess this is effectively `--force-but-ignore-bootsel`? :wink: 
I'm not sure that the `--only-force` name makes it clear what this is doing, so I'll have a ponder... :thinking: 

As a start, perhaps the help-text ought to be changed to "Same as --force, but will ignore devices in BOOTSEL mode" ?

main.cpp
will-v-pi: They should return `dr_vidpid_debugprobe_no_reset` and print the associated messages, as they have the same vid/pid, but no reset interface

picoboot_connection/picoboot_connection.h
will-v-pi: I need to think it through more and improve it - currently `dr_vidpid_usb_reset` means any device with the reset interface (stdio_usb or not), but `dr_vidpid_usb_reset_cant_connect` only applies to stdio_usb devices that cannot be opened (nothing to do with the reset interface), and I think devices without a USB reset interface will return `dr_vidpid_bootrom_no_interface` - so I need to add a separate return for stdio_usb with no reset interface, similar to the debug probe one, and change the can’t connect back to stdio_usb only (not applicable for the reset interface)

main.cpp
will-v-pi: This phrasing is used in a few places, so I just copied it here - I can change them all to “manually put into” if that is clearer

picoboot_connection/picoboot_connection.h
lurch: I'm glad that my awkward questions were useful :wink:

main.cpp
lurch: I think it is clearer, but perhaps @JeuneseRPi can offer a second opinion?

picoboot_connection/picoboot_connection.h
will-v-pi: I have added comments in `picoboot_connection.h` by all the enum values which should clarify what everything means, and re-arranged the messages in `main.cpp` to matcth the re-arranged order there. Hopefully that all makes sense, let me know if there are any issues

main.cpp
will-v-pi: I've reworded it - maybe `--force-ignore-bootsel` for the option, but that feels too long?

picoboot_connection/picoboot_connection.c
will-v-pi: I applied this, but actually it wasn't a typo - I will change it to `no filtering does not count as custom_vid_pid` to be clearer

PR #353, opened 2026-07-27
Display uf2 combine --offset default-value in hexadecimal

### LINKED ISSUE #352: uf2 combine: --offset representation
this was unexpected behaviour.

picotool v2.3.0: uf2 combine --offset produces different output for decimal and hexadecimal values although both represent the same offset.

Example:

picotool uf2 combine bootloader.uf2 app_a.uf2 test1.uf2 --offset 65536
picotool uf2 combine bootloader.uf2 app_a.uf2 test2.uf2 --offset 0x10000

test1.uf2 and test2.uf2 differ (cmp/sha256sum), and only the hexadecimal version boots correctly on RP2040.

lurch:
Possible fix for #352

### DIFF
--- a/README.md
+++ b/README.md
@@ -1067,7 +1067,7 @@ OPTIONS:
         --offset
             Offset second UF2 by amount
         <offset>
-            offset amount (hexadecimal, default to 0x0)
+            offset amount (hexadecimal; default 0x0)
     Partition options
         --partition
             Place second UF2 in partition (first UF2 must contain a partition table)
--- a/README.md
+++ b/README.md
@@ -1067,7 +1067,7 @@ OPTIONS:
         --offset
             Offset second UF2 by amount
         <offset>
-            offset amount (default to 0)
+            offset amount (default to 0x0)
     Partition options
         --partition
             Place second UF2 in partition (first UF2 must contain a partition table)
--- a/README.md
+++ b/README.md
@@ -1067,7 +1067,7 @@ OPTIONS:
         --offset
             Offset second UF2 by amount
         <offset>
-            offset amount (default to 0x0)
+            offset amount (hexadecimal, default to 0x0)
     Partition options
         --partition
             Place second UF2 in partition (first UF2 must contain a partition table)
--- a/README.md
+++ b/README.md
@@ -732,7 +732,7 @@ OPTIONS:
         --minor <minor>
             Add Minor Version
         --rollback <rollback> [<rows>..]
-            Add Rollback Version
+            Add Rollback Version, optionally specifying rollback rows in hex
     Configuration
         --hash
             Hash the file
@@ -966,7 +966,7 @@ OPTIONS:
         --abs-block
             Enforce support for an absolute block
         <abs_block_loc>
-            absolute block location (default to 0x10ffff00)
+            absolute block location (memory address; default 0x10ffff00)
 ```
 
 ## uf2
@@ -1021,7 +1021,7 @@ OPTIONS:
         --abs-block
             Add an absolute block
         <abs_block_loc>
-            absolute block location (default to 0x10ffff00)
+            absolute block location (memory address; default 0x10ffff00)
 ```
 
 ### combine
@@ -1077,7 +1077,7 @@ OPTIONS:
         --abs-block
             Add an absolute block
         <abs_block_loc>
-            absolute block location (default to 0x10ffff00)
+            absolute block location (memory address; default 0x10ffff00)
 ```
 
 The `--partition` argument can be used to place the second file in a partition number, provided that there is a partition table in the first file. For example, take this `pt.json`
@@ -1730,7 +1730,7 @@ OPTIONS:
         --verbose
             Print verbose output
         -p <pad>
-            Specify alignment to pad to, defaults to 0x1000
+            Specify alignment to pad to (hexadecimal; default 0x1000)
     File to write to
         <outfile>
             The file name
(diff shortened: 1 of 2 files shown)


main.cpp
```diff
@@ -1639,7 +1639,7 @@ struct uf2_combine_command : public cmd {
                 ).force_expand_help(true) % "UF2 Family options" +
                 (
                     option("--offset").set(settings.uf2.offset_set) % "Offset second UF2 by amount" &
-                        hex("offset").set(settings.uf2.offset) % "offset amount (default to 0)"
+                        hex("offset").set(settings.uf2.offset) % "offset amount (default to 0x0)"
```
will-v-pi: Maybe worth explicitly mentioning it is hexadecimal too?

```suggestion
                        hex("offset").set(settings.uf2.offset) % "offset amount (hexadecimal, default to 0x0)"
```

README.md
```diff
@@ -1067,7 +1067,7 @@ OPTIONS:
         --offset
             Offset second UF2 by amount
         <offset>
-            offset amount (default to 0)
+            offset amount (default to 0x0)
```
will-v-pi:
```suggestion
            offset amount (hexadecimal, default to 0x0)
```

main.cpp
lurch: Sorry, I somehow completely missed this comment! :facepalm: (must have got buried under all my other GH notifications)
I've no objections to this, but note that there's many other example / default values listed in https://github.com/raspberrypi/picotool/blob/develop/README.md that start with `0x` and that aren't explicitly stated to be hexadecimal.

will-v-pi: Most of those either say `memory address` (or similar), or are `integer()` command line options (which accept hex or non-hex values)

I think the only `hex()` options that don't explicitly say `memory address` or `hexadecimal` are:
- `--rollback` rows - could change to `Add Rollback Version, optionally specifying rollback rows in hex`
- `--pad` - could change to `Specify alignment to pad to (hexadecimal; default 0x1000)`
- `--abs-block` abs_block_loc - could change to `(memory address; default 0x10ffff00)`

lurch: Sounds good to me - feel free to make whatever changes you want to this PR :slightly_smiling_face:  (and then merge it when you're happy)

will-v-pi (approved):

README.md
```diff
@@ -732,7 +732,7 @@ OPTIONS:
         --minor <minor>
             Add Minor Version
         --rollback <rollback> [<rows>..]
-            Add Rollback Version
```
lurch, line 735: The rollback stuff is stored in OTP, right? So why is this using hex numbers when https://github.com/raspberrypi/picotool#setget appears to be using decimal numbers? (Or do the OTP docs also need to be updated to specify hexadecimal numbers? :shrug: )

will-v-pi: I'm not sure where the decimal row numbers are in the docs? Rows can be indexed by hex or decimal for `set`/`get`, but the output always prints hex:

```
$ picotool otp list 0x40
ROW 0x0040: OTP_DATA_CRIT1
        "Page 1 critical boot flags (RBIT-8)"
    field GLITCH_DETECTOR_SENS (bits 5-6)
    field GLITCH_DETECTOR_ENABLE (bit 4)
    field BOOT_ARCH (bit 3)
    field DEBUG_DISABLE (bit 2)
    field SECURE_DEBUG_DISABLE (bit 1)
    field SECURE_BOOT_ENABLE (bit 0)

$ picotool otp list 64
ROW 0x0040: OTP_DATA_CRIT1
        "Page 1 critical boot flags (RBIT-8)"
    field GLITCH_DETECTOR_SENS (bits 5-6)
    field GLITCH_DETECTOR_ENABLE (bit 4)
    field BOOT_ARCH (bit 3)
    field DEBUG_DISABLE (bit 2)
    field SECURE_DEBUG_DISABLE (bit 1)
    field SECURE_BOOT_ENABLE (bit 0
```

Same goes for `PAGE:PAGE_ROW_NUMBER` selectors - `1:0x11`, `1:17`, `0x1:17` etc all target the same row

This `--rollback` argument is using hex rows for simplicity, as the default rows are `OTP_DATA_DEFAULT_BOOT_VERSION0_ROW=0x4e` and `OTP_DATA_DEFAULT_BOOT_VERSION1_ROW=0x51`

lurch: > Rows can be indexed by hex or decimal

Ahhh okay, that's the bit I was missing. As it doesn't say "hexadecimal", that's why I was assuming it was decimal-only. But perhaps it's worth tweaking the OTP section of the docs to clarify that the row and page numbers can be specified in either decimal or hex?

will-v-pi: All regular `integer` arguments in `picotool` can be specified as hex or dec (or binary) (see `integer::parse_string` in `cli.h`, so I don't think that it is necessary to state that these selector fields also support hex or dec, given the output prints row numbers in hex anyway (for `set`/`get`/`list` and `dump`)

lurch: Is that actually mentioned in the README, or do people need to read the source-code to discover that? :wink:

will-v-pi: Not that I can see in the README, but anyway, I think that's a separate issue - this PR is about `hex` arguments which always treat values as hex (i.e. `1234` is treated as `0x1234`), whereas this discussion is about `integer` arguments which are treated as decimal unless a `0x` or `0b` prefix. The `hex` ones are the only ones with non-default behaviour, hence being good to have explicit documentation, `integer` behaves like you would expect it to behave.

lurch: > Not that I can see in the README, but anyway, I think that's a separate issue

Haha, fair enough, sorry for going off-topic.

PR #334, opened 2026-07-01
README improvements

will-v-pi:
Can view the new README at https://github.com/raspberrypi/picotool/tree/docs-links

Add links to each command near the top of the README

Reduce the line width to 100, so it fits without needing horizontal scrolling

PR #356, opened 2026-08-11
Fix two outdated facts in BUILDING.md

DanielMBouyou:
`CMakeLists.txt` compiles `DOCS_WIDTH=95`; BUILDING.md still says 140. #334 changed it deliberately ("Reduce command width in the README"), a month after that sentence was last touched.

Without libusb six commands go, not five: `bdev` is missing.

Not fixed: the README `otp permissions` example passes `--tx`, which is not an option: the only `--tx` in any tracked C/C++ source, ever, is the `--tx_remaining` decrement in flash_id.c. That block also has device output I can't regenerate.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -8,7 +8,7 @@ git submodule update --init lib/mbedtls
 
 You also need to install `libusb-1.0` if you want to use the USB functionality.
 
-> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot). Builds that do not include USB support can be identified by running `picotool version`, which will state `This version of picotool was compiled without USB support. Some commands are not available.`. Additionally, the unsupported commands won't appear in the output of the help command, and if you attempt to execute an invalid command you will get an error message. The build output message `libUSB is not found - no USB support will be built` also appears in the build logs.
+> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot, bdev). Builds that do not include USB support can be identified by running `picotool version`, which will state `This version of picotool was compiled without USB support. Some commands are not available.`. Additionally, the unsupported commands won't appear in the output of the help command, and if you attempt to execute an invalid command you will get an error message. The build output message `libUSB is not found - no USB support will be built` also appears in the build logs.
 
 ### Linux / macOS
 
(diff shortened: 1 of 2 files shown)


BUILDING.md
```diff
@@ -8,7 +8,7 @@ git submodule update --init lib/mbedtls
...
+> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, erase, verify, reboot, bdev). Builds that do not include USB support can be identified by running `picotool version`, which will state `This version of picotool was compiled without USB support. Some commands are not available.`. Additionally, the unsupported commands won't appear in the output of the help command, and if you attempt to execute an invalid command you will get an error message. The build output message `libUSB is not found - no USB support will be built` also appears in the build logs.
```
will-v-pi: I think adding etc would be better, as this was already not an exhaustive list

```suggestion
> If libusb-1.0 is not installed, picotool still builds, but it omits all options that deal with managing a pico via USB (load, save, reboot, etc.). Builds that do not include USB support can be identified by running `picotool version`, which will state `This version of picotool was compiled without USB support. Some commands are not available.`. Additionally, the unsupported commands won't appear in the output of the help command, and if you attempt to execute an invalid command you will get an error message. The build output message `libUSB is not found - no USB support will be built` also appears in the build logs.
```

will-v-pi:
Thanks, I've pushed a commit to this PR to address that final point about the `picotool otp permissions` output too

lurch:
> Not fixed: the README `otp permissions` example passes `--tx`, which is not an option

Curious. I wonder if that was from a previous version of https://github.com/raspberrypi/picotool#configuration ?

@will-v-pi On a side-note, https://github.com/raspberrypi/picotool#config doesn't explain what the `-s` flag stands for. And maybe it'd be nice if the `picotool config` examples showed a `bi_ptr_int32` value being modified, in addition to the current `bi_ptr_string` "name" example. It's also not clear how the "group" mentioned at https://github.com/raspberrypi/picotool#config relates to the https://github.com/raspberrypi/picotool#binary-information
(And I also just noticed that it talks about `bi_ptr_string` in a couple of places, but that's not included in the "bunch of bi_ macros" list at https://github.com/raspberrypi/picotool#details )

BUILDING.md
lurch: "a pico" should probably also be "a Pico".

will-v-pi:
> > Not fixed: the README `otp permissions` example passes `--tx`, which is not an option
> 
> Curious. I wonder if that was from a previous version of https://github.com/raspberrypi/picotool#configuration ?

It's from the internal RP2350 development pre-2.0.0 - the binary originally printed UART output, hence having a configurable TX pin, but that was removed before the 2.0.0 release and replaced with just an LED.

> @will-v-pi On a side-note, https://github.com/raspberrypi/picotool#config doesn't explain what the `-s` flag stands for. And maybe it'd be nice if the `picotool config` examples showed a `bi_ptr_int32` value being modified, in addition to the current `bi_ptr_string` "name" example. It's also not clear how the "group" mentioned at https://github.com/raspberrypi/picotool#config relates to the https://github.com/raspberrypi/picotool#binary-information (And I also just noticed that it talks about `bi_ptr_string` in a couple of places, but that's not included in the "bunch of bi_ macros" list at https://github.com/raspberrypi/picotool#details )

It stands for `set` (as in `set <key> to <value>`) - that is already fixed as part of #342, along with better documentation of `group`. This is just a quick demo, there is a much more complete example in [pico-examples](https://github.com/raspberrypi/pico-examples/tree/master/binary_info), so I don't think more needs adding here.

DanielMBouyou:
Both applied.

You're right the list wasn't exhaustive — I'd measured six commands dropping without libusb rather than five, so `etc.` is sturdier than any list I'd have written. Capitalised Pico on that line too.

Thanks for pushing the `--tx` fix, and for the pre-2.0.0 explanation — that answers it.

will-v-pi (approved):

PR #359, opened 2026-09-18
Refactor integer & hex to be more common

will-v-pi:
Allows deriving family_id from new integer_base with hex_format

Also fixes some edge cases (e.g. 64-bit partition IDs)


cli.h
```diff
@@ -418,212 +418,188 @@ namespace cli {
...
-            if (value.find("0x") == 0) {
+            auto ovalue = value;
+            if (!std::is_signed<T>() && value.find('-') == 0) {
+                // stoull would silently wrap this round
+                return ovalue + " must not be negative";
+            }
+            int64_t base = FORMAT::radix;
+            if ((!base || base == 16) && value.find("0x") == 0) {
                 value = value.substr(2);
                 base = 16;
```
lurch, line 483: I have to confess I don't fully follow what this code is doing, but it _looks_ like in the old version, if `base` was 10 and `value` started with `0x`, then it'd update `base` to 16; but in the new version if `base` was 10 and `value` started with `0x`, then it'd leave `base` set to 10?

(Hmmm, I wonder if this was supposed to be `(!base || base != 16)` ? :thinking: )

lurch: Also looks a bit weird that an `int64_t` is used to store the radix, but perhaps that just means there's less type-conversions needed in later calculations?

EDIT: Ahh, [`std::stoll`](https://en.cppreference.com/cpp/string/basic_string/stol) and [`std::stoull`](https://en.cppreference.com/cpp/string/basic_string/stoul) both declare `base` as `int`, but clarify that "The set of valid values for `base` is `{0, 2, 3, ..., 36}`" :slightly_smiling_face:

will-v-pi: This is intentional - if `base` is already set it should not remove any `0x` prefix, unless `base` is already set to 16. This is so that in the future, if an option required a base 10 input it would reject anything with a `0x` prefix (although no options do that at the moment).

cli.h
```diff
@@ -418,212 +418,188 @@ namespace cli {
         }
     };
 
-    struct integer : public value_base<integer> {
-        explicit integer(string name) : value_base(std::move(name)) {}
+    // Number format policies for integer_base; they say how a value is parsed, and how
+    // values are worded/rendered in error messages
+    struct decimal_format {
+        static constexpr int64_t radix = 0; // 0 means work the base out from the 0x/0b prefix
```
lurch: Following on from the previous comments - as `hex_format` is base-16-only, I'd inadvertently assumed that `decimal_format` is base-10-only. But I now see that this is actually "base 0" i.e. infer-the-base-from-the-prefix, so I wonder if `guess_format` might be a better name than `decimal_format`?

cli.h
```diff
@@ -418,212 +418,188 @@ namespace cli {
...
+        static constexpr uint64_t value_mask = std::numeric_limits<uint64_t>::max();
+        static const char *name() { return "integer"; }
+        static const char *too_big_error() { return " is too big"; }
+        static string format(int64_t value) { return std::to_string(value); }
+    };
+
+    struct hex_format {
+        static constexpr int64_t radix = 16; // always base 16; an 0x prefix is optional
+        static constexpr uint64_t value_mask = 0xffffffffu; // 32 bit values only
```
lurch, line 433: Are we ever likely to need to support a 64-bit `hex_format` in the future? :thinking:

cli.h
will-v-pi: I've gone with `integer_format`, to match the `integer` and `integer_base`, as it is the default for those

cli.h
will-v-pi: Unlikely - `hex_format` is mostly used for memory addresses, and the RP2 processors are all 32-bit (also family IDs, which are 32-bit too)

lurch: Fair enough :+1: 
(I was only asking because your PR comment says "(e.g. 64-bit partition IDs)", but maybe partition IDs will always be `integer_format` rather than `hex_format` :slightly_smiling_face: )

